Issue #44, opened 2020-07-05
Configurable linux capabilities (cap_add / cap_drop)


towe75:
We should give users the chance to add/drop linux capabilities.

see https://www.nomadproject.io/docs/drivers/docker#cap_add

PR #45, opened 2020-07-05
#44 Configurable linux capabilities (cap_add / cap_drop)

towe75:


driver_test.go
```diff
@@ -1029,3 +1060,42 @@ func getPodmanDriver(t *testing.T, harness *dtestutil.DriverHarness) *Driver {
...
+	defer d.DestroyTask(task.ID, true)
+
+	// Attempt to wait
+	waitCh, err := d.WaitTask(context.Background(), task.ID)
+	require.NoError(t, err)
+
+	select {
+	case <-waitCh:
+		t.Fatalf("wait channel should not have received an exit result")
+	case <-time.After(time.Duration(tu.TestMultiplier()*2) * time.Second):
```
drewbailey, line 1096: this seems like we will always wait for 2 seconds? maybe instead we can bring over or use `testutil.WaitForResult` https://github.com/hashicorp/nomad/blob/master/testutil/wait.go#L17

README.md
```diff
@@ -252,6 +252,26 @@ config {
...
+    "SYS_TIME"
+  ]
+}
+```
+
+* **cap_drop** - (Optional)  A list of Linux capabilities as strings to pass to --cap-drop.
+
+```
+config {
+  cap_add = [
```
drewbailey, line 269: cap_drop

PR #49, opened 2020-07-16
deps: update GitHub action used Golang version to 1.14.5.

jrasell:


towe75:
It seems like GH actions automatically chooses "latest" go from some local cache. 

> It will first check the local cache for a version match. If version is not found locally, It will pull it from main branch of go-versions repository and on miss or failure, it will fall back to the previous behavior of download directly from go dist.


Today i merged a PR and triggered a build:

```
Run actions/setup-go@v2
Setup go stable version spec ^1.14.1
Found in cache @ /opt/hostedtoolcache/go/1.15.0/x64
Added go to the path
Successfully setup go version ^1.14.1
go version go1.15 linux/amd64
```

setup-go installed go1.15 (.. Found in cache ...).
But compilation failed:

```
Run sudo -E ./build.sh
go: downloading github.com/varlink/go v0.3.0
# internal/unsafeheader
compile: version "go1.15" does not match go tool version "go1.14.7"
# internal/cpu
compile: version "go1.15" does not match go tool version "go1.14.7"
# runtime/internal/atomic
compile: version "go1.15" does not match go tool version "go1.14.7"
# runtime/internal/sys
compile: version "go1.15" does not match go tool version "go1.14.7"
...
...
```

Obviously there is some path clash between setup-go binary (v1.5) and some other go runtime (1.14.7). 
Maybe the ubuntu base image includes it?

We should ensure to not have some "rogue" go binary in the base image.
Also the behavior is not really predictable due to the unknown state of the builders local cache. I propose to pin the go-version by removing the "^".

towe75:
The [ubuntu virtual environment 18.04](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md) includes indeed multiple (!) go binaries. 
One of those is marked as the default, currently it's the newest 1.14.x version. See [https://github.com/LKaemmerling/virtual-environments/blob/main/images/linux/toolsets/toolset-1804.json](tool spec)

```
        {
            "name": "go",
            "url" : "https://raw.githubusercontent.com/actions/go-versions/main/versions-manifest.json",
            "arch": "x64",
            "platform" : "linux",
            "versions": [
                "1.11.*",
                "1.12.*",
                "1.13.*",
                "1.14.*"
            ],
            "default": "1.14.*"
        },
```

So maybe we can simple remove the setup-go action and stick to the included default version?

towe75:
@jrasell @drewbailey do we want to keep this actions/setup-go or should we rely on the ubuntu virtual environment. 

I removed actions/setup-go in branches [here ](https://github.com/hashicorp/nomad-driver-podman/pull/68/commits/df274feb2e15968bfaeb0da1567d701d1421dad2) and [here](https://github.com/hashicorp/nomad-driver-podman/blob/b2cb7a33b3f163c99717a082f455fa396b3af27f/.github/workflows/build.yml) and did not see issues.

towe75:
@jrasell @drewbailey : can we close and remove the branch? We're already building against go1.15.10 .

PR #54, opened 2020-07-31
config: Ability to configure dns server list

towe75:
Driver config can now specify a list of dns servers:

* **dns** - (Optional)  A list of dns servers. Replaces the default from podman binary and containers.conf.

```
config {
  dns = [
    "1.1.1.1"
  ]
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+FEATURES:
+
+* config: Ability to configure dns server list
+
 ## 0.1.0
 
 FEATURES:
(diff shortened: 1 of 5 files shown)


drewbailey:
Looks good to me!

drewbailey:
Hey @towe75 I missed that this Nomad PR had landed https://github.com/hashicorp/nomad/pull/7661 I think we should instead allow the value set in `network > dns` to be plummed through to podman

drewbailey:
Created https://github.com/hashicorp/nomad-driver-podman/issues/56 to track

PR #58, opened 2020-08-14
retry getting inspect container info before failing

drewbailey:


driver.go
```diff
@@ -518,9 +518,22 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
...
+		inspectData, inspectErr = d.podmanClient.InspectContainer(containerID)
+		if err != nil {
+			d.logger.Debug("failed to inspect container, retrying", "err", err)
+			continue
+		}
+		break
+	}
+
+	if inspectErr != nil {
+		d.logger.Error("failed to inspect container, retrying", "err", err)
```
jrasell, line 536:
```suggestion
		d.logger.Error("failed to inspect container", "err", err)
```

drewbailey: Woops thanks :)

towe75:
@drewbailey: i tried the WaitContainer approach in my http-apiclient branch and it works very well.

drewbailey:
@towe75 that sounds more promising than this change then!

towe75:
@drewbailey  the api V2 code can be found [here](https://github.com/hashicorp/nomad-driver-podman/blob/b2cb7a33b3f163c99717a082f455fa396b3af27f/apiclient/container_start.go#L46). 

I tried to fix this issue in 0.1.x using varlinks "WaitContainer()" api call. But it does not allow us to specify a state, rather "stop" is hardcoded, so i can not use it. 

What to make out of this branch? Do you still want to polish and merge it or should we go ahead and focus on API V2?

drewbailey:
No sense in trying to get this in, let's focus on v2

PR #80, opened 2021-01-03
WIP: actor based state handling, streaming container stats and podman events

towe75:
Goals of this branch:

* avoid mutexes for taskhandle storage
* replace per-task container stats polling with a single stats stream
* replace per-task container dead check with a podman event stream observer

General idea is to replace the mutex based task store with a message based actor. The actors mailbox, a simple channel, serializes all requests and makes explicit locking unnecessary in a lot of cases.


apollo13:
Hi @towe75, this looks great. What would be needed to move this forward? Also I See that you are restarting the stats/event streams after some delay if the fail. Wouldn't that mean that we loose events?

PR #83, opened 2021-01-15
config: Support for sysctl configuration (#82)

### LINKED ISSUE #82: Can we add --sysctl in podman driver?
Hi,

    Sometimes you need to set some kernel values like net.core.somaxconn with sysctl. 
    It is available in docker driver, can we also have it for podman driver?

Thanks!

towe75:
Fixes #82


towe75:
@drewbailey PTAL. It's a small change and we can merge/release it after your review.

drewbailey (approved):
lgtm!

PR #85, opened 2021-01-20
add exec support

drewbailey:
Creating a Draft PR containing WIP for basic exec support.

I don't think we'll easily be able to support TTY without significant work in handling / upgrading connections to websockets, we may be able to shim the docker go api package to use their exec command against a podman endpoint

### DIFF
--- a/api/exec.go
+++ b/api/exec.go
@@ -1,0 +1,49 @@
+package api
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+
+	docker "github.com/docker/docker/api/types"
+)
+
+type ExecCreateConfig struct {
+	docker.ExecConfig
+}
+
+type ExecCreateResponse struct {
+	docker.IDResponse
+}
+
+type ExecStartConfig struct {
+	Detach bool `json:"Detach"`
+	Tty    bool `json:"Tty"`
+}
+
+func (c *API) CreateExec(ctx context.Context, container string, opts ExecCreateConfig) (*ExecCreateResponse, error) {
+	payload, err := json.Marshal(opts)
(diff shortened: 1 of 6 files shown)


towe75:
I checked some code and found that both websocket and raw tcp can be used:

* cockpit-podman has a http based remote terminal using a websocket: https://github.com/cockpit-project/cockpit-podman/blob/master/src/ContainerTerminal.jsx
* podmans own binding package upgrades the http call to a raw tcp connection. Here we would get distinct stderr/stdout streams but handling is somewhat complex: https://github.com/containers/podman/blob/v2.1/pkg/bindings/containers/attach.go

Maybe we can borrow/share some code from there?
Documentation is rather poor, i could not find any details whatsoever.

towe75:
@drewbailey could not resist to give it a try. Please take a look at my branch https://github.com/hashicorp/nomad-driver-podman/commit/55d442078d9f32fad07fc9b2a46e7feb7acfa66f

I ported podmans own attach.go to fit into our client structure. I've got a working interactive cli (nomad alloc exec) and web-shell, also scripted consul health checks are and using e.g. pipelines in a shell is also ok. 

For sure it needs some polishing and something is still wrong with terminal/window resizing - although i did the necessary plumbing and neither nomad nor podman complain about something. 

I am also not sure if stderr forwarding works correctly in interactive mode, never tried the consul client so i don't know how it should behave. 

It's certainly a very good starting point and it might even be enough for master.
Be aware that it must be adopted to my actor branch later.

towe75:
Resize problem seems to be a podman bug, see https://github.com/containers/podman/issues/7102

towe75:
@drewbailey : can we close this one in favor of #87 ?

PR #86, opened 2021-01-28
config: added tty option

towe75:


towe75:
@krishicks PTAL. Adds yet another config option.

krishicks (approved):

PR #87, opened 2021-01-30
Exec support using libpod api

towe75:
Alternative implementation to #85
It uses libpod api directly, following the minimal client strategy of our 0.2.0 release.

It's pretty complete and support both interactive use and scripted health checks. The only known issue is related to terminal resize events due to https://github.com/containers/podman/issues/7102 which should be fixed in the upcoming podman 3.0 release.

### DIFF
--- a/api/exec_create.go
+++ b/api/exec_create.go
@@ -1,0 +1,85 @@
+package api
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+// ExecConfig contains the configuration of an exec session
+type ExecConfig struct {
+	// Command the the command that will be invoked in the exec session.
+	// Must not be empty.
+	Command []string `json:"Cmd"`
+	// DetachKeys are keys that will be used to detach from the exec
+	// session.
+	DetachKeys string `json:"DetachKeys,omitempty"`
+	// Environment is a set of environment variables that will be set for
+	// the first process started by the exec session.
+	Environment map[string]string `json:"Env,omitempty"`
+	// The user, and optionally, group to run the exec process inside the container.
+	// Format is one of: user, user:group, uid, or uid:gid."
+	User string `json:"User,omitempty"`
+	// WorkDir is the working directory for the first process that will be
+	// launched by the exec session.
+	// If set to "" the exec session will be started in / within the
+	// container.
+	WorkDir string `json:"WorkingDir,omitempty"`
+	// Tty is whether the exec session will allocate a pseudoterminal.
+	Tty bool `json:"Tty,omitempty"`
+	// AttachStdin is whether the STDIN stream will be forwarded to the exec
+	// session's first process when attaching. Only available if Terminal is
+	// false.
+	AttachStdin bool `json:"AttachStdin,omitempty"`
+	// AttachStdout is whether the STDOUT stream will be forwarded to the
+	// exec session's first process when attaching. Only available if
+	// Terminal is false.
+	AttachStdout bool `json:"AttachStdout,omitempty"`
+	// AttachStderr is whether the STDERR stream will be forwarded to the
+	// exec session's first process when attaching. Only available if
+	// Terminal is false.
+	AttachStderr bool `json:"AttachStderr,omitempty"`
+	// Privileged is whether the exec session will be privileged - that is,
+	// will be granted additional capabilities.
+	Privileged bool `json:"Privileged,omitempty"`
+}
+
+//
+type ExecSessionResponse struct {
+	Id string
+}
+
+// ExecCreate creates an exec session to run a command inside a running container
+func (c *API) ExecCreate(ctx context.Context, name string, config ExecConfig) (string, error) {
+
+	jsonString, err := json.Marshal(config)
+	if err != nil {
+		return "", err
+	}
+
+	res, err := c.Post(ctx, fmt.Sprintf("/v1.0.0/libpod/containers/%s/exec", name), bytes.NewBuffer(jsonString))
+	if err != nil {
+		return "", err
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusCreated {
+		body, _ := ioutil.ReadAll(res.Body)
+		return "", fmt.Errorf("unknown error, status code: %d: %s", res.StatusCode, body)
+	}
+
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return "", err
+	}
+	execResponse := &ExecSessionResponse{}
+	err = json.Unmarshal(body, execResponse)
+	if err != nil {
+		return "", err
+	}
+	return execResponse.Id, err
+}
--- a/api/exec_create.go
+++ b/api/exec_create.go
@@ -47,9 +47,9 @@ type ExecConfig struct {
 	Privileged bool `json:"Privileged,omitempty"`
 }
 
-//
+// ExecSessionResponse contains the ID of a newly created exec session
 type ExecSessionResponse struct {
-	Id string
+	ID string
 }
 
 // ExecCreate creates an exec session to run a command inside a running container
@@ -81,5 +81,5 @@ func (c *API) ExecCreate(ctx context.Context, name string, config ExecConfig) (s
 	if err != nil {
 		return "", err
 	}
-	return execResponse.Id, err
+	return execResponse.ID, err
 }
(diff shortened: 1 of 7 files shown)


towe75:
@drewbailey  PTAL. I think it's low risk to merge because it adds a totally new feature without changing any of the existing functions.

api/exec_create.go
```diff
@@ -0,0 +1,85 @@
...
+	// AttachStderr is whether the STDERR stream will be forwarded to the
+	// exec session's first process when attaching. Only available if
+	// Terminal is false.
+	AttachStderr bool `json:"AttachStderr,omitempty"`
+	// Privileged is whether the exec session will be privileged - that is,
+	// will be granted additional capabilities.
+	Privileged bool `json:"Privileged,omitempty"`
+}
+
+//
```
drewbailey, line 50:
```suggestion
// ExecSessionResponse contains the ID of a newly created exec session
```

api/exec_create.go
```diff
@@ -0,0 +1,85 @@
...
+type ExecSessionResponse struct {
+	Id string
```
drewbailey, line 52: Nit but I think ID is preferred 
```suggestion
	ID string
```

api/exec_start.go
```diff
@@ -0,0 +1,217 @@
...
+	}()
+
+	buffer := make([]byte, 1024)
+	if options.Tty {
+		// If not multiplex'ed, read from server and write to stdout
+		_, err := io.Copy(options.Stdout, socket)
+		if err != nil {
+			return err
+		}
+	} else {
```
drewbailey, line 172: Can we return nil after the copy and drop this else

api/exec_start.go
```diff
@@ -0,0 +1,217 @@
...
+			if rerr != nil {
+				c.logger.Warn("failed to resize TTY", "err", rerr)
+			}
+		}
+	}
+}
+
+// ExecStartAndAttach starts and attaches to a given exec session.
+func (c *API) ExecStart(ctx context.Context, sessionID string, options ExecStartRequest) error {
+	client := *c.httpClient
```
drewbailey, line 101: having our own clone function may be helpful here, but for now I think something like the following would be a bit safer
```
	client := new(http.Client)
	*client = *c.httpClient
```

api/exec_start.go
```diff
@@ -0,0 +1,217 @@
...
+// This is intended to be run as a goroutine, handling resizing for a container
+// or exec session.
+func (c *API) attachHandleResize(ctx context.Context, resizeChannel <-chan drivers.TerminalSize, sessionId string) {
+	for {
+		select {
+		case <-ctx.Done():
+			c.logger.Warn("Resize handler is done")
+			return
+		case size := <-resizeChannel:
+			c.logger.Debug("Resize terminal", "sessionId", sessionId, "height", size.Height, "width", size.Width)
```
drewbailey, line 90:
```suggestion
			c.logger.Trace("Resize terminal", "sessionId", sessionId, "height", size.Height, "width", size.Width)
```

driver.go
```diff
@@ -700,9 +701,107 @@ func (d *Driver) SignalTask(taskID string, signal string) error {
...
+		AttachStderr: true,
+	}
+	ctx, cancel := context.WithTimeout(d.ctx, timeout)
+	defer cancel()
+	sessionId, err := d.podman.ExecCreate(ctx, handle.containerID, createRequest)
+	if err != nil {
+		d.logger.Error("Unable to create ExecTask session", "err", err)
+		return nil, err
+	}
+	stdout, _ := circbuf.NewBuffer(int64(drivers.CheckBufSize))
```
drewbailey, line 724: let's handle these error cases

drewbailey (commented):
I tried this out locally and it works great! Excellent work :tada: Just a few comments, Some tests would be good too

api/exec_start.go
towe75: options.Tty is false for non interactive exec sessions, used by consul health checks. I tried to mimic what the docker driver does and the original podman code had this if/else case to support it.
We can try to simplify but it will require some testing

api/exec_start.go
towe75: I'm fine with the change because i am not really a go expert ;-)

towe75:
@drewbailey : i rebased this branch to include the latest changes from main. Please check my earlier reply on your review.

Are we good to go? I think it would be worth a release.

api/exec_start.go
```diff
@@ -0,0 +1,218 @@
...
+	return buffer[0:length], nil
+}
+
+// This is intended to be run as a goroutine, handling resizing for a container
+// or exec session.
+func (c *API) attachHandleResize(ctx context.Context, resizeChannel <-chan drivers.TerminalSize, sessionId string) {
+	for {
+		select {
+		case <-ctx.Done():
+			c.logger.Warn("Resize handler is done")
```
drewbailey, line 88:
```suggestion
			c.logger.Trace("Resize handler is done")
```

api/exec_start.go
```diff
@@ -0,0 +1,218 @@
...
+			return
+		case size := <-resizeChannel:
+			c.logger.Trace("Resize terminal", "sessionId", sessionId, "height", size.Height, "width", size.Width)
+			rerr := c.ExecResize(ctx, sessionId, size.Height, size.Width)
+			if rerr != nil {
+				c.logger.Warn("failed to resize TTY", "err", rerr)
```
drewbailey, line 94:
```suggestion
				c.logger.Error("failed to resize TTY", "err", rerr)
```

drewbailey:
@towe75 this looks really great! A few tests would be great to ensure everything works as expected as new podman versions roll out. There should be a few good examples in the Nomad repo

towe75:
@drewbailey : thank you for feedback. I'll try to add some tests.

drewbailey:
@towe75 We can also merge this and do a follow up PR with some tests before releasing if you prefer, I'll approve the PR and leave it up to you. If we can fix those two log levels that would be great

drewbailey (approved):

api/exec_start.go
towe75: done

towe75:
@drewbailey i started to derive a test from the other task driver and only two sub-tests from StreamingConformingTests fail:

--- FAIL: TestPodmanDriver_execTaskStreaming (76.20s)
    --- PASS: TestPodmanDriver_execTaskStreaming/basic:_notty:_basic (0.13s)
    --- PASS: TestPodmanDriver_execTaskStreaming/basic:_notty:_streaming (3.12s)
    --- PASS: TestPodmanDriver_execTaskStreaming/basic:_notty:_stty_check (0.11s)
    --- PASS: TestPodmanDriver_execTaskStreaming/basic:_notty:_stdin_passing (0.12s)
    --- FAIL: TestPodmanDriver_execTaskStreaming/basic:_notty:_children_processes (1.09s)
    --- PASS: TestPodmanDriver_execTaskStreaming/basic:_tty:_basic (0.13s)
    --- PASS: TestPodmanDriver_execTaskStreaming/basic:_tty:_streaming (3.07s)
    --- PASS: TestPodmanDriver_execTaskStreaming/basic:_tty:_stty_check (1.10s)
    --- PASS: TestPodmanDriver_execTaskStreaming/basic:_tty:_stdin_passing (0.11s)
    --- PASS: TestPodmanDriver_execTaskStreaming/basic:_tty:_children_processes (1.10s)
    --- FAIL: TestPodmanDriver_execTaskStreaming/isolation (0.28s)

notty:children_processes has to be tackled, could be potential bug. But "isolation" test has a hardcoded docker assumption in exec_testing.go . Should this become a bug in the nomad repo?

I did not commit the test yet to not break the build.

drewbailey:
@towe75 that is great news! sounds like `TestExecFSIsolation` should pass in the expected freezer cgroup instead of assuming nomad or docker

towe75:
I looked into the other failing test and it seems to be a podman problem OR wrong assumption. 
Here is what i tried:

```
podman run -d -t busybox --name tester

podman exec tester /bin/sh -c "(( sleep 3; echo from background ) & ); echo from main; exec sleep 1"
from main
```

A plain run without a container acts differently:

```
/bin/sh -c "(( sleep 3; echo from background ) & ); echo from main; exec sleep 1"
from main
from background
```

Should i raise a podman issue? Seems they handle it differently compared to docker.

drewbailey:
@towe75 fyi we are removing this test https://github.com/hashicorp/nomad/pull/10291 from nomad

towe75:
@drewbailey thank you for keeping me posted. I will file a follow-up issue to add the tests later because there is still the TestExecFSIsolation blocker. 
After that we can merge this PR, IMHO.

PR #88, opened 2021-02-13
Fixed a bug where we always pulled an image if image name had a transport prefix

towe75:
Images prefixed with a transport where always pulled.
I noticed this while examining podman logs during unit test runs. It re-pulled busybox for each single test.

This PR improves image url parsing and also makes the code a bit nicer by adding an extra method for image handling.

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -24,6 +24,7 @@ import (
 	"github.com/hashicorp/nomad/plugins/shared/hclspec"
 	pstructs "github.com/hashicorp/nomad/plugins/shared/structs"
 
+	dockerref "github.com/docker/distribution/reference"
 	shelpers "github.com/hashicorp/nomad/helper/stats"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 )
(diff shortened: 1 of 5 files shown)


towe75:
@krishicks : please review, this also fixes #90 when using podman 2.x

driver.go
```diff
@@ -583,6 +573,57 @@ func memoryInBytes(strmem string) (int64, error) {
...
+		tag = tagged.Tag()
+	}
+
+	digested, ok := named.(dockerref.Digested)
+	if ok {
+		digest = digested.Digest().String()
+	}
+	if len(tag) == 0 && len(digest) == 0 {
+		tag = "latest"
+	}
```
krishicks, line 607: It might be nice to extract all of this into its own function, since most of this function exists to manipulate a given image. That could then have its own set of tests to make it clear what the expected output for a given input are.

At that point I'd remove the `ensureImage` function entirely, as it would be about the same length as it was before.

krishicks (approved):
This looks good to me, though I would like to see the image-transforming code extracted into a separate, tested function.

driver.go
towe75: @krishicks PTAL, i added a separate, tested, function. I think it's ok to also keep ensureImage() because of possible, future, improvements regarding oci-archive etc.

krishicks: I like the new tested function. I would still remove `ensureImage`, as any future change that warrants bringing it back can do so at that time. Otherwise, it's YAGNI.

My approval still applies, in any event.

Issue #89, opened 2021-02-14
Any plans to support podman 3.0.0


chris-rock:
[Podman 3.0.0](https://github.com/containers/podman/releases/tag/v3.0.0) has just landed and I was wondering if nomad is going to support the latest version?

towe75:
Hi @chris-rock , thank you for reaching out.

Tests with podman 3.0 rc where successful and but it seems that a last minute API change is a show-stopper. See https://github.com/containers/podman/issues/9351

Also there are various ubuntu related setup/upgrade problems which is worse given the fact that only 3.0 is available in their repository, see https://github.com/containers/podman/issues/9358 . For now we're waiting for a point release to see if it improves the API situation.  Otherwise we can, of course, react with some version checking and if/else API adapters.

chris-rock:
Thank you for the quick feedback. I using alpine edge and it already ships the latest podman 3 as a package.

matejvasek:
@chris-rock @towe75 The incompatible change was introduced accidentally. It should be fixed in `v3.0.1`.

towe75:
@matejvasek : thank you for keeping us posted!

chris-rock:
That is great. Thank you @matejvasek @towe75 Once podman 3.0.1 is alpine, I am going to do another try...

towe75:
@matejvasek : 3.0.1 is out and i checked if our plugin "just works". I can see that the podman gh issue is resolved but it's a long, convoluted thread and i can not see if the API request works now as planed.

Fact is that i still have to adopt our codebase. We're using "/v1.0.0/libpod/containers/%s/wait?condition=%s" but only "/v1.0.0/libpod/containers/%s/wait?condition[]=%s" (note the brackets!) seems to do the job. Should i open another issue or is it what you want to do in this API? It's a bit annoying to add the brackets even if i only wait on one state.

matejvasek:
the brackets break functionality for me: `http://127.0.0.1:1234/v1.0.0/libpod/containers/festive_poitras/wait?condition[]=running` immediately returns on non running container.

matejvasek:
how did you figure to put it there?

matejvasek:
I think that `condition[]` is discarded == it waits for default condition which is stopped.

matejvasek:
I see only call is `err = c.ContainerWait(timeout, name, "running")` right?

towe75:
@matejvasek : my bad, sorry. I ran into a combination of:

[...]

PR #91, opened 2021-02-24
Improve container_start to improve startup of short living containers

towe75:
Improve container startup state tracking. Some details can be found in #89 
It helps to overcome a race condition when running very short living containers in e.g. unit tests.

A better solution would be to continue on #80


towe75:
@drewbailey : PTAL . it fixes a flaky test. After that all our tests pass with podman 3.0.1 (see GH action log)

drewbailey (approved):
lgtm!

PR #94, opened 2021-03-21
Adopt some rootless code to podman 3.0.x

### LINKED ISSUE #92: Driver is not reading correct rootless info
Hi guys, recently I have been playing with the driver quite some and found some information displayed in nomad ui incorrectly, which is the `rootless` attribute.

Per the current implementation, it's reading from `info.Host.Rootless`: https://github.com/hashicorp/nomad-driver-podman/blob/bf21e157fc5124fbce0a44cae2f6a608b6cd466f/driver.go#L224
However I think it should be `host.security.rootless` per the podman's documentation:  https://docs.podman.io/en/latest/_static/api.html#operation/libpodGetInfo

Please let me know if I got things wrong, and thank you for this awesome driver!

### LINKED ISSUE #93: Possible discrepancy between README and code regarding default network mode for rootless containers
Hi,

The README says that `slirp4netns` is the default network mode for rootless containers but that doesn't seem to be the case when testing and looking at the code. It seems the default is `bridge` even when running rootless containers. I stumbled into this when playing around with Nomad for the very first time so there's plenty of room for mistakes on my part.

Anyway my solution was to simply specify `network_mode = "slirp4netns"` in the config-section. This was tested on v0.2.0 release and 0502b79 (with bf21e157 reverted as it breaks with Podman 2.2.1), and I'd be happy to help test any patches.

Nomad v1.0.4
Podman 2.2.1

Documentation:
https://github.com/hashicorp/nomad-driver-podman/blame/main/README.md#L240
Code:
https://github.com/hashicorp/nomad-driver-podman/blob/0502b790bcafe0b87ea77290559b0c259586895c/driver.go#L433

Thanks,
Andreas

towe75:
fix #92 parse rootless info correctly from podman 3.0.x struct
fix #93 use slirp4netns as default network mode if running rootless
Bump supported podman version in README to 3.0.x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 * config: Support for sysctl configuration [[GH-82](https://github.com/hashicorp/nomad-driver-podman/issues/82)]
 * config: Fixed a bug where we always pulled an image if image name has a transport prefix [[GH-88](https://github.com/hashicorp/nomad-driver-podman/pull/88)]
 
+BUG FIXES:
+* [[GH-93](https://github.com/hashicorp/nomad-driver-podman/issues/93)] use slirp4netns as default network mode if running rootless
+* [[GH-92](https://github.com/hashicorp/nomad-driver-podman/issues/92)] parse rootless info correctly from podman 3.0.x struct
+
 ## 0.2.0
 
 FEATURES:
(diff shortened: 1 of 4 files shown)


towe75:
@drewbailey : please have a look. It fixes two rootless problems.

drewbailey (approved):
lgtm!

PR #96, opened 2021-04-04
Set tag when pulling image missing in local storage

pablofsf:
First of all, thank you very much for this plugin!

I was testing some basic configuration with nomad and this plugin and bumped into the following issue: 

With a clean podman installation, I tried to run the following task, which failed with `task=apache error="rpc error: code = Unknown desc = failed to start task, could not create container: unknown error, status code: 500: {"cause":"no such image","message":"unable to find 'docker.io/httpd:2' in local storage: no such image","response":500}"`
```
task "apache" {
  driver = "podman"

  config {
    image = "docker.io/httpd:2"
    ports = ["http"]
  }
}
```

To my surprise, the image saved to storage was `httpd:latest` instead of `httpd:2`. I understand the reason is that podman interprets the missing tag as `latest` when asking `ImagePull` just for the `imageName`

This PR fixes this issue and adds tag information to the logging and error in case the image is not present in local storage. I also understand that this should be the expected behavior according to the docker driver documentation.

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -507,6 +507,7 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 				return nil, nil, fmt.Errorf("failed to start task, unable to pull image %s : %v", imageName, err)
 			}
 		}
+		createOpts.Image = imageName
 
 		createResponse, err := d.podman.ContainerCreate(d.ctx, createOpts)
 		for _, w := range createResponse.Warnings {
--- a/driver.go
+++ b/driver.go
@@ -488,23 +488,23 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 		//        e.g. oci-archive:/... etc
 		//        see also https://github.com/hashicorp/nomad-driver-podman/issues/69
 
-		imageName, tag, err := parseImage(createOpts.Image)
+		imageName, err := parseImage(createOpts.Image)
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to start task, unable to parse image reference %s: %v", createOpts.Image, err)
 		}
 
 		// do we already have this image in local storage?
-		haveImage, err := d.podman.ImageExists(d.ctx, fmt.Sprintf("%s:%s", imageName, tag))
+		haveImage, err := d.podman.ImageExists(d.ctx, imageName)
 		if err != nil {
 			d.logger.Warn("Unable to check for local image", "image", imageName, "err", err)
 			// do NOT fail this operation, instead try to pull the image
 			haveImage = false
 		}
 		if !haveImage {
-			d.logger.Debug("Pull image", "image", fmt.Sprintf("%s:%s", imageName, tag))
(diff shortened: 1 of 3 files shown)


towe75:
@pablofsf thank you for this PR, it's very much appreciated! The bug is a regression of #88 and it would be nice to cleanup a bit more. 

There are now several identical Sprintf(name,tag) invocations in consecutive lines. It would be better to format the string once and use it multiple times. Also the actual createOpts struct does not rely on the parsed and formatted name but still uses the "raw" driverConfig.Image.

Would you mind to invest a few minutes more to clean it up?

pablofsf:
Absolutely. I'll check out the mentioned PR and clean it up a bit. 

Would it maybe make sense to also add a test to catch a possible future regression?

pablofsf:
@towe75 should be done now. I have added a test to avoid the regression from happening again. I have tested locally that just running the test without the changes in the first commit would fail with the same result as in the description. However, I though squashing the test and the fixed would make more sense than adding a broken commit to the history, even if it makes it harder for verification.

towe75:
@pablofsf : i found that your PR did not trigger the GH action workflow to run the tests. This problem was not noticed before because most changes still happens through pushes instead of "external" PRs. 

Can you please add the pull_request trigger to https://github.com/hashicorp/nomad-driver-podman/blob/d3c2186ac8b46300050c18d3adff5a39aaacfb7b/.github/workflows/build.yml#L3 within your repo? This should kick off a test-run.

driver.go
```diff
@@ -507,6 +507,7 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 				return nil, nil, fmt.Errorf("failed to start task, unable to pull image %s : %v", imageName, err)
 			}
 		}
+		createOpts.Image = imageName
```
pablofsf, line 510: @towe75 I added this to fix an error with the digest. I understand that the issue is that podman treats differently images with tags and digests. Whereas tagged images are considered shortnames if there is not a fully-qualified reference, images with digests are not treated like that: https://github.com/containers/image/blob/v5.10.5/pkg/shortnames/shortnames.go#L17

Also,  the fact that `parseImage` uses `dockerref.ParseNormalizedNamed` deviates from standard podman behavior. It will insert `docker.io` by default if a domain is missing, completely ignoring podman shortname behavior. However, I think fixing that is out of the scope if this PR and deserves a separate discussion.

towe75: Yes, IC and in fact i proposed this change in my first reply to your email: 

> Also the actual createOpts struct does not rely on the parsed and formatted name but still uses the "raw" driverConfig.Image.

It's not so simple to fix it in a clean way until there is better tooling/library-support for podman available.

towe75 (approved):

towe75:
@pablofsf : LGTM, thank you! @drewbailey: do you want to take a look as well?

drewbailey (approved):
LGTM, thank you!

driver.go
pablofsf: OK, perfect. Sorry I did not realize about that until it failed. I will open an issue to track this, might have the time to look into it in the following weeks.

PR #98, opened 2021-04-06
Fix errors missed while pulling images

pablofsf:
On the /images/pull endpoint, podman's StatusOK response does
not warranty that the image was actually pulled, just that
the request was well formatted and no InteralServerError happened.
If any error happens during the download, the error is written to
the output stream in the error field.

I bumped into this by doing a typo in an image name. The driver failed creating the container complaining about a missing image, which in theory had been pulled but was not in the container store. Manually requesting the API for the image yielded this result:
```
{"stream":"Trying to pull docker.io/library/nextcloud:alpine...\n"}
{"stream":"  manifest unknown: manifest unknown\n"}
{"error":"Error initializing source docker://nextcloud:alpine: Error reading manifest alpine in docker.io/library/nextcloud: manifest unknown: manifest unknown\n"}
```
This PR should fix that problem.

I think ideally a test should be added to check this behavior at the API level. However, as no API method has a test, I wondered whethere there is a specific reason for that.

### DIFF
--- a/api/image_pull.go
+++ b/api/image_pull.go
@@ -2,7 +2,10 @@ package api
 
 import (
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 )
@@ -16,10 +19,23 @@ func (c *API) ImagePull(ctx context.Context, nameWithTag string) error {
 	}
 
 	defer res.Body.Close()
-	body, _ := ioutil.ReadAll(res.Body)
-
 	if res.StatusCode != http.StatusOK {
+		body, _ := ioutil.ReadAll(res.Body)
 		return fmt.Errorf("unknown error, status code: %d: %s", res.StatusCode, body)
+	}
+
+	dec := json.NewDecoder(res.Body)
+	var report ImagePullReport
+	for {
+		if err = dec.Decode(&report); err == io.EOF {
+			break
+		} else if err != nil {
+			return fmt.Errorf("Error reading response: %w", err)
+		}
+
+		if report.Error != "" {
+			return errors.New(report.Error)
+		}
 	}
 
 	return nil
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
         path: nomad-driver-podman
 
     - name: Test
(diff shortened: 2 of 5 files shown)


towe75:
@pablofsf thank you again for the research and the fix. I will do a review later on.

Their reference at https://docs.podman.io/en/latest/_static/api.html#operation/ImagePullLibpod indeed specifies a response body. I might have overlooked this in the initial implementation.  It's not very REST'isch, though. A simple 404 would certainly be nicer. Anyhow, good you found it.

Their is no specific reason for missing api tests and contributions are welcome ;-)

pablofsf:
> @pablofsf thank you again for the research and the fix. I will do a review later on.

Thank you for all your work in this plugin! I am in the process to move to nomad+podman and currently doing a lot of testing. So if contributions are welcome, you can expect some more more PR for fixes and features to get this in sync with the docker plugin :)

> Their reference at https://docs.podman.io/en/latest/_static/api.html#operation/ImagePullLibpod indeed specifies a response body. I might have overlooked this in the initial implementation. It's not very REST'isch, though. A simple 404 would certainly be nicer. Anyhow, good you found it.

Yes, it is strange. From all the methods we use, it seems to be the only one that does something like that. ContainerWait also replies with an exit code, but that is just an integer and probably not of much interest.

> Their is no specific reason for missing api tests and contributions are welcome ;-)

Ok, perfect. I've added a commit with the test before the fix, so it is a bit more obvious than it is actually fixing something. Could squash them if that is preferred

api/image_pull.go
```diff
@@ -16,11 +19,24 @@ func (c *API) ImagePull(ctx context.Context, nameWithTag string) error {
...
 		return fmt.Errorf("unknown error, status code: %d: %s", res.StatusCode, body)
 	}
 
+	dec := json.NewDecoder(res.Body)
+	var report ImagePullReport
+	for {
+		if err = dec.Decode(&report); err == io.EOF {
+			break
+		} else if err != nil {
+			return fmt.Errorf("Error reading response: %v", err)
```
towe75, line 33: @pablofsf please use %w for new code to wrap the error.

towe75 (changes requested):

api/image_pull.go
pablofsf: Fixed!

towe75 (approved):

towe75:
@pablofsf: Awesome! Thank you

Issue #69, opened 2020-11-06
Run containers from oci-archive and non standard image sources


towe75:
It would be useful to run containers from images in oci-archive format.
We fixed this issue for varlink api in #67 

But switching to podman api v2 introduces the same problem. This is related to https://github.com/containers/podman/issues/6744

onlyjob:
Looks like 0.2.0 was actually released with this regression... :(

CC: @drewbailey, @krishicks

towe75:
@onlyjob : it was a misleading information in the changelog,  #67 was fixed right before 0.2.0 release. 
I knew that it would still be broken with http API - hence this ticket. I will handle it soon and this time it's covered with a unit test ;-)

onlyjob:
I've backported fix to 0.1.0 and it works with Varlink. However yesterday I tried 0.2.0 and could not start any of my images...

Thank you for your care and attention for driver-podman.

towe75:
We're blocked by https://github.com/containers/podman/issues/8927

towe75:
@onlyjob do you follow the podman issue? For now there is no way to use the http API to get it working. What is your opinion?

onlyjob:
Yes, I'm following, thanks... Not upgrading is my only option, obviously... :(  But that also means that we can't disable deprecated Varlink in Debian's Podman...

pablofsf:
@onlyjob, this has in principle been fixed in #101. There is a simple test which worked successfully and imports the right images from the archives. However, I did not do any real-world testing. If you are able to check it out, maybe there is a chance any further bugs appear or we can consider this as solved. Only `oci-archive` and `docker-archive` are currently allowed transports other than the default `docker`

PR #101, opened 2021-04-11
Honor podman shortnames behavior and allow non-docker transports

### LINKED ISSUE #97: Do not rely on docker.io for non fully-qualified references
Currently, due to the use of `dockerref.ParseNormalizedNamed` in `parseImage`, `docker.io` will be prepended by default if the image does not contain a domain.
https://github.com/hashicorp/nomad-driver-podman/blob/9e1733f3b49007d211fa337f3041e73589eae1dd/driver.go#L619

This deviates from podman behavior for [shortnames](https://www.redhat.com/sysadmin/container-image-short-names) and could also result in unexpected behavior if the user does not have `docker.io` in the first position of `unqualified-search-registries`.

pablofsf:
Closes #97
It also attempts to tackle #69

So this PR does two things:
* First, identified whether the image provided is a shortname, in which case it will be dealt with by podman according to user configuration.
* Second, adds support for `oci-archive` and `docker-archive`. Would this be enough for #69? podman also supports `dir` and `docker-daemon` transports, but I believe those would not work with `podman load`, although I have not tested it. 

Unfortunately by using `containers/image` there are quite some dependencies pulled. I believe I could reimplement/copy some of those methods to limit the dependency list if needed.


driver_test.go
```diff
@@ -1381,9 +1381,39 @@ func startDestroyInspectImage(t *testing.T, image string, taskName string) {
...
+
+func Test_createImage(t *testing.T) {
+	if !tu.IsCI() {
+		t.Parallel()
+	}
+
+	d := podmanDriverHarness(t, nil)
+	// Need to test: shortname, longname with/without transport
+	// TODO: Also test oci/docker archives
+	// TODO: Make sure that images do NOT exist before some tests
```
pablofsf, line 1399: This two TODO probably require adding  `api.ImageRemove` and `api.ImageSave` methods. Would it make sense to add them even if they are only used for the tests?

towe75:
@pablofsf  thank you for another contribution. I will have a look in next couple of days.

pablofsf:
Rebased to fix the merge conflicts. @towe75, is there any way I could help to ease the review?

towe75:
@pablofsf : thank you for rebasing it. 
No, the review only depends on available developer capacity right now ;-)

towe75 (changes requested):
@pablofsf : nice work!
Only drawback is the amount of dependencies it pulls. Please have a look at the remaining conversation. Also i want to wait for a second review by somebody else because of the impact of this change.

api/image_load.go
```diff
@@ -0,0 +1,39 @@
...
+		return "", err
+	}
+	if res.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("unknown error, status code: %d: %s", res.StatusCode, body)
+	}
+	err = json.Unmarshal(body, &names)
+	if err != nil {
+		return "", err
+	}
+	return names[0], nil
```
drewbailey, line 38: If for whatever reason names is empty this will panic, do you mind adding a length check just to be safe?

driver.go
```diff
@@ -632,35 +618,85 @@ func memoryInBytes(strmem string) (int64, error) {
...
+// returns the 64-byte image ID as an unique image identifier
+func (d *Driver) createImage(image string) (string, error) {
+	var imageID string
+	imageName := image
+	// If it is a shortname, we should not have to worry
+	// Let podman deal with it according to user configuration
+	if !shortnames.IsShortName(image) {
+		imageRef, err := parseImage(image)
+		if err != nil {
+			return imageID, fmt.Errorf("invalid image reference %s: %w\n", image, err)
```
drewbailey, line 631: Why the new line?
```suggestion
			return imageID, fmt.Errorf("invalid image reference %s: %w", image, err)
```

driver.go
```diff
@@ -632,35 +618,85 @@ func memoryInBytes(strmem string) (int64, error) {
...
 		}
+		// If its non-docker transport, we cannot ask for a pull or
+		// check for existence in the API without image plumbing
+		if imageRef.Transport().Name() != "docker" {
+			if imageRef.Transport().Name() != "oci-archive" &&
+				imageRef.Transport().Name() != "docker-archive" {
+				return imageID, fmt.Errorf("unssupported transport: %s", imageRef.Transport().Name())
+			}
```
drewbailey, line 639: This was a bit hard for me to parse, do you mind condensing it into a switch to handle the error instead?
```suggestion
		switch name := imageRef.Transport().Name(); name {
		case "docker", "oci-archive", "docker-archive":
		default:
			return imageID, fmt.Errorf("unssupported transport: %s", imageRef.Transport().Name())
		}

```

driver.go
```diff
@@ -632,35 +618,85 @@ func memoryInBytes(strmem string) (int64, error) {
...
+			// docker-archive:path[:docker-reference]
+			// oci-archive:path:tag
+			path := strings.Split(image, ":")[1]
```
drewbailey, line 642: The value for `image` is user supplied, right? If so, I think we should length check to avoid a panic

driver.go
```diff
@@ -632,35 +618,85 @@ func memoryInBytes(strmem string) (int64, error) {
...
+		}
+		imageName = imageRef.DockerReference().String()
 	}
 
-	named, err := dockerref.ParseNormalizedNamed(image)
+	imageID, err := d.podman.ImageInspectID(d.ctx, imageName)
 	if err != nil {
-		return "", err
+		d.logger.Warn("Unable to check for local image", "image", imageName, "err", err)
+		// do NOT fail this operation, instead try to pull the
```
drewbailey, line 656:
```suggestion
		// If ImageInspectID errors, continue the operation and try to pull
		// the image instead
		d.logger.Warn("Unable to check for local image", "image", imageName, "err", err)
```

driver.go
```diff
@@ -632,35 +618,85 @@ func memoryInBytes(strmem string) (int64, error) {
...
+	case "docker":
+		return docker.ParseReference(name)
+	case "oci-archive":
+		return ociArchive.ParseReference(name)
+	case "docker-archive":
+		return dockerArchive.ParseReference(name)
+	default:
+		// We could have both an unknown/malformed transport
+		// or an image:tag with default "docker://" transport omitted
+		ref, err := docker.ParseReference("//" + image)
```
drewbailey, line 694: Isn't this handled above in the `len(parts) == 1` check? I think we can just error here instead.

drewbailey (commented):
@pablofsf Thank you for your contribution! This is an exciting improvement. I just have a few comments and questions, thanks again!

driver.go
pablofsf: The check is not needed because the image already went through `parseImage`. Therefore, we know for sure that in the case in which the transports are identified as `oci-archive` or `docker-archive`, the image was properly parsed by `ociArchive.ParseReference` or `dockerArchive.ParseReference`. Thus, it should be safe. 

However, I understand it is not so intuitive. The other possible solution would be:
```suggestion
			archiveData := imageRef.StringWithinTransport()
			path := strings.Split(archiveData, ":")[0]
```
Although I need to finish the archive tests first to verify this works as intended.

drewbailey: Ah gotcha, missed that this morning, makes sense! I do like the suggestion you provided

driver.go
pablofsf: This is me sometimes thinking in C by default...

api/image_load.go
pablofsf: Thank you for this one! I added a check and some error message.

driver.go
pablofsf: Actually, the tests failed due to this, so I reverted it. The problem that we have here is in case somebody sends an image like `docker.io/library/busybox:latest` expecting the docker transport to be used as default. As the `:` is the same character used to separated the transport to the image to the tag, we have no efficient way to identify that in `parts`. 

I've changed the error handling a bit to reflect this a bit better. Is it more understandable now?

driver.go
pablofsf: Thank you for this suggestion! It looks so much nicer as a switch.

I've changed it a bit to remove the default case. We shouldn't deal with unsupported transports here but in `parseImage`.

driver.go
pablofsf: It's fixed now

pablofsf:
@drewbailey, I've added two tests and modified the pipeline as suggested by @towe75. I had to do one fix related to the image load, so I've added an extra commit to ease the review, but I can squash them all once it's approved. 

However, the pipeline is not running. Maybe it detects that I'm not a collaborator and the changes need to be approved? Otherwise I guess I might have unintentionally broken it... 

In any case, tests run fine locally, so I guess this should cover most of issue #69. Is there anything else that would be required to close that issue? Maybe adding the list of supported transports to the documentation would be useful for users?

And thank you both for your reviews!

drewbailey:
@pablofsf docs would be great, but can also be in a follow up PR if you'd like. I'm not sure why the GH action isn't running, likely due to contributor permissions. We can always merge in and fix anything that pops up. Approving, thanks again for your contribution!

drewbailey (approved):

.github/workflows/build.yml
```diff
@@ -1,11 +1,14 @@
...
 
-on: 
+on:
   pull_request:
     branches:
       - main
   workflow_dispatch:
 
+env:
+  ARCHIVE_DIR=/tmp
```
towe75, line 10: @pablofsf : this is invalid syntax and the reason for the broken test pipeline. Please compare the file with latest master. It was already improved to handle pull_requests. Maybe you can simply revert your change. Otherwise please fix this line.

towe75 (approved):

.github/workflows/build.yml
pablofsf: Thanks for the pointer, it's fixed now! The archive test also run successfully in the pipeline

towe75:
Seems we're good to go! @drewbailey : ok to merge now?

PR #78, opened 2020-12-27
Issue #56 Plumb through Nomad network -> dns to podman

towe75:
Fixes #56

PR #102, opened 2021-04-13
Cleanup unused dns in taskConfigSpec

pablofsf:
Small nit. I guess this was left from #78


towe75 (approved):

towe75:
@pablofsf : good catch, thank you

PR #104, opened 2021-04-15
Handle memory oversubscription by setting hard/soft limits

notnoop:
The complexity is handling `reservation`, a driver field that sets soft
limits. I opted to honor current behavior, and still use it as the soft
limit if running in oversubscription mode.


drewbailey (approved):
lgtm!

PR #106, opened 2021-04-17
Update github action runner to ubuntu 20.04 and Podman 3.1.x

towe75:
[Podman team announced ](https://podman.io/blogs/2021/03/02/podman-support-for-older-distros.html) that they will no longer build packages for Ubuntu 18.04. 

This PR upgrades our github action runner to Ubuntu 20.04 and this brings us also from podman 3.0.1 to podman 3.1.x.

### DIFF
--- a/.github/machinesetup.sh
+++ b/.github/machinesetup.sh
@@ -1,8 +1,8 @@
 #!/bin/bash -e
 
-# add podman 2.x repository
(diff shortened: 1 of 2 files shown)


drewbailey (approved):

drewbailey:
we should make a note to update the local vagrant environment. I'll need to figure out what to do  with the nomad e2e test suite that runs a podman job, that currently uses 18.04 🤦

drewbailey:
Ah seems like we have until podman 4.x to deal with nomad e2e and docs. That could also be in a month or two for all I know :D 

> Podman 3.0 will be the last major build on CentOS 7, Debian 10 and Ubuntu 18.04.

Issue #71, opened 2020-12-02
Images from a private registry / auth


mmack:
It would be great to support the auth flag like the docker driver does: https://www.nomadproject.io/docs/drivers/docker#authentication

rodericliu:
I have been testing this plugin, eveything works fine except the registry authentication. It would be great to have this.

towe75:
Yes, this is a important feature. We will try to add it soon.

towe75:
I checked our options. So far i found that the http API supports only very simple username/password authentication: https://docs.podman.io/en/latest/_static/api.html#operation/libpodImagesPull . I can imagine that we rather want credential helper support for AWS ECR and that alike.

On the other hand: console allows to use credential helpers and docker compatible auth files: https://www.mankier.com/5/containers-auth.json

What's your opinions? Should we start with a simple username/password option and grow it later once there is better API available? It's quite possible that we could use the http API with credential helpers and that only some documentation is missing.

towe75:
It's indeed not possible to configure credential helpers via http API. This means that i will only implement a simple username/password authentication at this time.

elimisteve:
@towe75 Sounds great to me! 👍

towe75:
Implemented via #112

kholisrag:
can we use this method to auth from AWS ECR private repo?

PR #112, opened 2021-05-16
config: Image registry authentication #71

towe75:
This addresses #71 
It adds a auth {username, password} task configuration option.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -108,7 +108,7 @@ func (c *API) Delete(ctx context.Context, path string) (*http.Response, error) {
 	return c.Do(req)
 }
 
-// encode auth configuration to a docker'isch X-Registry-Auth header payload
+// NewAuthHeader encodes auth configuration to a docker X-Registry-Auth header payload.
 func NewAuthHeader(auth ImageAuthConfig) (string, error) {
 	jsonBytes, err := json.Marshal(auth)
 	if err != nil {
--- a/api/api.go
+++ b/api/api.go
@@ -2,6 +2,8 @@ package api
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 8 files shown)


api/api.go
```diff
@@ -93,3 +107,13 @@ func (c *API) Delete(ctx context.Context, path string) (*http.Response, error) {
 	}
 	return c.Do(req)
 }
+
+// encode auth configuration to a docker'isch X-Registry-Auth header payload
```
drewbailey, line 111:
```suggestion
// NewAuthHeader encodes auth configuration to a docker X-Registry-Auth header payload.
```

api/image_pull.go
```diff
@@ -11,10 +11,17 @@ import (
 )
 
 // ImagePull pulls a image from a remote location to local storage
-func (c *API) ImagePull(ctx context.Context, nameWithTag string) (string, error) {
+func (c *API) ImagePull(ctx context.Context, nameWithTag string, auth ImageAuthConfig) (string, error) {
 	var id string
 
-	res, err := c.Post(ctx, fmt.Sprintf("/v1.0.0/libpod/images/pull?reference=%s", nameWithTag), nil)
+	authHeader, err := NewAuthHeader(auth)
```
drewbailey, line 17: I think we may want to check if the supplied `ImageAuthConfig` is the zero value or use a pointer, so we can check if auth is empty and conditionally set it.

If ImageAuthConfig is an empty struct right now I think it will get json encoded to `{}` and then base64 encoded to `e30=`, which may be confusing or potentially result in an error. https://play.golang.org/p/xHknDDMTl8A

drewbailey (commented):
Looks good to me! I think we just want to only set the request header values if ImageAuthConfig is actually set

api/image_pull.go
towe75: done

drewbailey (approved):
lgtm!

PR #113, opened 2021-05-23
Fix network_mode=container:$id network option

towe75:
README explains that we can use network_mode = container:$othercontainer to let one container joint the network of another, already running, container. This feature was not really implemented.

This PR adds now the missing implementation, fixes a podman json structure and includes a unit test.


drewbailey (approved):
nice catch, thanks for adding the test

PR #114, opened 2021-05-27
Added support for using existing network stack

joshiparth1000:
Support for using existing network stack

### DIFF
--- a/api/container_create.go
+++ b/api/container_create.go
@@ -55,6 +55,11 @@ type ContainerCreateRequest struct {
 	// Optional.
 	Command []string `json:"command,omitempty"`
 
+	// Pod is the container's pod.
+	// If not given and container will not be inside a pod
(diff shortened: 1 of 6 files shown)


towe75:
@joshiparth1000 thank you for your contribution!
I will take a closer look within a few days. Maybe you can check PR #116 in the meantime to see if there is some feature overlap with your work? Both branches enhance the drivers network interoperability.

towe75:
@joshiparth1000 : What about integrating your feature with the existing network_mode driver configuration option. It would build upon the ideas from #116, adding a network_mode=pod:somepod and network_mode=cni:first,second

joshiparth1000:
Yes, that will work as well. I am up for it.

shoenig:
As mentioned in #173 Podman is dropping CNI support in v5.

PR #115, opened 2021-05-28
Improved StopTask in case that container was lost outside nomad

towe75:
Handle stopped and unknown/lost containers correctly while stopping a task.
The previous version let nomad run circles around lost containers. This happend when a group was reallocated to another node when a host was rebooted. After rebook, nomad tried to stop the already stopped container etc.  
I found that the docker driver has a similar if/else chain in StopTask, so i reassembled it here.

### DIFF
--- a/api/container_stop.go
+++ b/api/container_stop.go
@@ -10,14 +10,18 @@ import (
 // timeout value.  The timeout value the time before a forcible stop to the container is applied.
 // If the container cannot be found, a [ContainerNotFound](#ContainerNotFound)
 // error will be returned instead.
-func (c *API) ContainerStop(ctx context.Context, name string, timeout int) error {
+func (c *API) ContainerStop(ctx context.Context, name string, timeout int, ignoreStopped bool) error {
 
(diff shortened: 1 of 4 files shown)


drewbailey (approved):
looks good to me!

towe75:
@drewbailey : please check again. I changed the code because of some comments in #116 . Now, we're simply using the existing "ignore" option in the podman container-stop API call.

api/container_stop.go
```diff
@@ -10,15 +10,19 @@ import (
 // timeout value.  The timeout value the time before a forcible stop to the container is applied.
 // If the container cannot be found, a [ContainerNotFound](#ContainerNotFound)
 // error will be returned instead.
-func (c *API) ContainerStop(ctx context.Context, name string, timeout int) error {
+func (c *API) ContainerStop(ctx context.Context, name string, timeout int, ignoreStopped bool) error {
```
drewbailey, line 13: Not necessary for now, but this seems like something we may eventually want to support a generic QueryOptions struct for as the list of options grows

drewbailey (approved):

api/container_stop.go
towe75: Yep, let's see if it grows again.

PR #116, opened 2021-05-28
Add flexible network namespaces and shared group network namespace support for consul connect

towe75:
This PR add several new network related features.

First of all we can have now shared group network namespaces for e.g. consul connect. I aimed for rootful support and i am not sure if a rootless container can be joined to an existing network namespace somehow. 
A rootful container can, without any problems, join the nomad controlled network ns. We do not need a separate "pause" container like the docker driver implements it. This also means that this driver does not have to implement the CreateNetwork/DestroyNetwork hooks, so i it is more lightweight compared to the docker driver.

Additionally we have now two new network_mode values:

* network_mode = ns:/some/network/namespace  
* network_mode = task:name-of-other-task-in-same-group

The "task:" variant is very useful because a user can build arbitrary pod-like structures with it. The key is to combine lifecycle stanzas and the new mode. A prestart/sidecar task becomes the network namespace (e.g. pause container) and the main workload can reference it easily. Nice is that other tasks in the same group are not forced to join the namespace, they can still have their own mode.

- [ ] write more tests
- [x] write more documentation
- [x] add a pod example 

## Examples

examples/jobs/nats_pod.nomad shows how to run a pod with a pause, main-workload and exporter sidecar, all in a common network namespace:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,54 @@ this plugin to Nomad!
 * Fine tune memory usage: standard [Nomad memory resource](https://www.nomadproject.io/docs/job-specification/resources.html#memory) plus additional driver specific swap, swappiness and reservation parameters, OOM handling
 * Supports rootless containers with cgroup V2
 * Set DNS servers, searchlist and options via [Nomad dns parameters](https://www.nomadproject.io/docs/job-specification/network#dns-parameters)
+* Support for nomad shared network namespaces and consul connect
+* Quite flexible [network configuration](network-configuration), allows to simply build pod-like structures within a nomad group
 
+## Redis Example job
+
+Here is a simple redis "hello world" Example:
+
+```
+job "redis" {
+  datacenters = ["dc1"]
+  type        = "service"
+
+  group "redis" {
+    network {
+      port "redis" { to = 6379 }
+    }
+
+    task "redis" {
+      driver = "podman"
+
+        config {
+          image = "docker://redis"
+          ports = ["redis"]
+        }
+
+      resources {
+        cpu    = 500
+        memory = 256
+      }
+    }
+  }
+}
+```
+
+```sh
+nomad run redis.nomad
+
+==> Monitoring evaluation "9fc25b88"
+    Evaluation triggered by job "redis"
+    Allocation "60fdc69b" created: node "f6bccd6d", group "redis"
+    Evaluation status changed: "pending" -> "complete"
+==> Evaluation "9fc25b88" finished with status "complete"
+
+podman ps
+
+CONTAINER ID  IMAGE                           COMMAND               CREATED         STATUS             PORTS  NAMES
+6d2d700cbce6  docker.io/library/redis:latest  docker-entrypoint...  16 seconds ago  Up 16 seconds ago         redis-60fdc69b-65cb-8ece-8554-df49321b3462
+```
 
 ## Building The Driver from source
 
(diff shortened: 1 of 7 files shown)


towe75:
@drewbailey : this surely becomes a great feature - although it was actually very simple to accomplish . Maybe you find the time to comment my work ?

driver.go
```diff
@@ -728,8 +742,24 @@ func (d *Driver) StopTask(taskID string, timeout time.Duration, signal string) e
 	if !ok {
 		return drivers.ErrTaskNotFound
 	}
+	_, err := d.podman.ContainerStats(d.ctx, handle.containerID)
```
drewbailey, line 745: @towe75 what was the motivation for this check? If the container stats response is in a state we did not anticipate is there any harm in trying to stop it anyway?

driver.go
```diff
@@ -456,6 +463,13 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 		} else if strings.HasPrefix(driverConfig.NetworkMode, "container:") {
 			createOpts.ContainerNetworkConfig.NetNS.NSMode = api.FromContainer
 			createOpts.ContainerNetworkConfig.NetNS.Value = strings.TrimPrefix(driverConfig.NetworkMode, "container:")
+		} else if strings.HasPrefix(driverConfig.NetworkMode, "ns:") {
+			createOpts.ContainerNetworkConfig.NetNS.NSMode = api.Path
+			createOpts.ContainerNetworkConfig.NetNS.Value = strings.TrimPrefix(driverConfig.NetworkMode, "ns:")
+		} else if strings.HasPrefix(driverConfig.NetworkMode, "task:") {
+			createOpts.ContainerNetworkConfig.NetNS.NSMode = api.FromContainer
+			otherContainerName := fmt.Sprintf("%s-%s", strings.TrimPrefix(driverConfig.NetworkMode, "task:"), cfg.AllocID)
```
nickethier, line 471: Can we create a shared method to derive a container name from task name+allocID? I worry that should we ever need to change how the name is built it could be whack a mole trying to find all the Sprintf instances (a game I've had the displeasure to play).

driver.go
nickethier: :+1: to @drewbailey's question. I'd think if the container ended up in an unexpected state before we stop it, a user would want to know.

driver_test.go
```diff
@@ -1270,7 +1270,7 @@ func TestPodmanDriver_NetworkModes(t *testing.T) {
 	}
 }
 
-// let a task joint NetorkNS of another container
+// let a task joint NetorkNS of another container via network_mode=container:
```
nickethier, line 1273:
```suggestion
// let a task join NetworkNS of another container via network_mode=container:
```

driver_test.go
```diff
@@ -1343,6 +1343,79 @@ func TestPodmanDriver_NetworkMode_Container(t *testing.T) {
 	require.Contains(t, tasklog, "127.0.0.1:6748")
 }
 
+// let a task joint NetorkNS of another container via network_mode=task:
```
nickethier, line 1346:
```suggestion
// let a task join NetworkNS of another container via network_mode=task:
```

nickethier (commented):
This is a cool addition!

driver.go
towe75: Oh, this change is actually part of #115 and should not have ended up here.

driver.go
towe75: @nickethier : sure. I was lazy but you caught me ;-)

towe75: I added a tool function, ok now?

towe75:
@drewbailey , @nickethier i added nice examples and tried to explain the feature in the readme. I wanted more tests but i am unsure how to write one for shared network namespaces (nomad "bridge" networking). The "network_mode = task:xyz" is already covered.

I think it's ready, please review it.

README.md
```diff
@@ -1,4 +1,4 @@
-Nomad podman Driver
+iomad podman Driver
```
drewbailey, line 1:
```suggestion
Nomad podman Driver
```

driver.go
drewbailey: 👍 if you don't mind just backing it out and taking care of it in #115 that works, or we can merge #115 and rebase :)

drewbailey (commented):
looking good, one small typo

README.md
towe75: Funny, a classic vim typo

towe75:
@drewbailey : i removed the leaked commit, fixed the readme typo and resolved merge commits. Are we GTG?

drewbailey (approved):
lgtm!

PR #119, opened 2021-06-04
Add labels option

zyclonite:


README.md
```diff
@@ -293,6 +293,17 @@ config {
 
 * **tty** - (Optional)  true or false (default). Allocate a pseudo-TTY for the container.
 
+* **labels** - (Optional)  Set labels on the container.
+
+```
+config {
+  options = [
```
towe75, line 300: @zyclonite : should be "labels".

config.go
```diff
@@ -53,6 +53,7 @@ var (
 		"image":              hclspec.NewAttr("image", "string", true),
 		"init":               hclspec.NewAttr("init", "bool", false),
 		"init_path":          hclspec.NewAttr("init_path", "string", false),
+		"labels":             hclspec.NewAttr("labels", "list(map(string))", false),
```
towe75, line 56: Any reason why it's not just a map(string) ?

towe75 (changes requested):

README.md
zyclonite: ups, you are right

config.go
zyclonite: i did look that up from the docker driver... i was not sure as well, never used hcl before

zyclonite: https://github.com/hashicorp/nomad/blob/bdf2555b385f547cbf613d1636eac103a1c58ce8/drivers/docker/config.go#L361

towe75: IC. But the docker config example shows no "[" , "]" list elements, just a map. Can you try to run you code against a simplified hcl without the brackets?

zyclonite: sure, works as well

towe75:
@zyclonite the hcl test and examples look certainly nicer now, thank you. 

Maybe you can write a simple driver_test, too? Check the TestPodmanDriver_Tty for a good example. It boils down to see if the attached labels are visible in the inspected container. After that we're good to go.

zyclonite:
@towe75 driver test added

towe75 (approved):
@zyclonite  LGTM!

PR #122, opened 2021-06-09
config: correctly handle multiple mount options per bind mount

towe75:
We did not handle multiple mount options correctly. A comma separated string was not split up but instead handed over as a single option, thus configuring the container failed. 

A test is added to proof that it is working correctly now.


jrasell (approved):
LGTM!

PR #125, opened 2021-06-17
handle: remove commented out log line.

jrasell:


towe75 (approved):
Ah, one from the early days. Sure!

PR #126, opened 2021-06-17
chore: use error context to log errors rather than Go err style.

jrasell:


towe75 (approved):
I'm fine with it. What is the exact difference if i use "error" here? Is it some hclog feature or just a hashicorp style convention?

jrasell:
> I'm fine with it. What is the exact difference if i use "error" here? Is it some hclog feature or just a hashicorp style convention?

Nothing to do with hclog functionality and mostly to do with consistency. When utilising centralised log aggregation we want to ensure search and filter patterns are as easy as possible for operators. Allowing a single search of "error" across Nomad core and drivers means less overhead when creating error tracking panels. Additionally, the use of "err" is a idiomatic Go and doesn't have a common wider use.

PR #130, opened 2021-08-09
Respect telemetry.collection_interval to reduce cpu churn when running many containers 

towe75:
Use telemetry.collection_interval for each container monitor. Previously we polled for each container with a very fast, static, 1 second interval. This does not scale very well and we can see a considerable load on the podman service when running e.g. a job with 30 pause containers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ FEATURES:
 
 BUG FIXES:
 * log: Use error key context to log errors rather than Go err style. [[GH-126](https://github.com/hashicorp/nomad-driver-podman/pull/126)]
+* telemetry: respect telemetry.collection_interval to reduce cpu churn when running many containers [[GH-130](https://github.com/hashicorp/nomad-driver-podman/pull/130)]
 
 ## 0.3.0
 
--- a/handle.go
+++ b/handle.go
@@ -28,6 +28,8 @@ type TaskHandle struct {
 	userCPUStats   *stats.CpuStats
 	systemCPUStats *stats.CpuStats
 
+	collectionInterval time.Duration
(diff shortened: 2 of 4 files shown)


handle.go
```diff
@@ -28,6 +28,8 @@ type TaskHandle struct {
 	userCPUStats   *stats.CpuStats
 	systemCPUStats *stats.CpuStats
 
+	collectionInterval time.Duration
```
lgfa29, line 31: Is there a specific why the driver would need to store the collection interval? The value would be passed from Nomad based on the client config, so it seems very local to `runStatsEmitter`.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ FEATURES:
 
 BUG FIXES:
 * log: Use error key context to log errors rather than Go err style. [[GH-126](https://github.com/hashicorp/nomad-driver-podman/pull/126)]
+* respect telemetry.collection_interval to reduce cpu churn when running many containers
```
lgfa29, line 8:
```suggestion
* telemtery: respect telemetry.collection_interval to reduce cpu churn when running many containers [[GH-130](https://github.com/hashicorp/nomad-driver-podman/pull/130)]
```

lgfa29 (approved):
Nice catch! Minor question about storing the need to store the interval, but definitely not a blocker.

handle.go
lgfa29: Ahh that makes sense, thanks! I missed noticing that it was crossing function boundaries 😅

ttys3:
before this I found my podman socket daemon flooded by health check log.

I have to increase podman log level to `error`

PR #131, opened 2021-08-24
http client timeout config

### LINKED ISSUE #100: Pulling images fails after 1 minute due to Timeout
I realized while having a look at #99, that the current timeout on `api.Post` for any api operation has a timeout of just one minute. That is definitely enough for most operations, but can easily become an issue for pulls in systems with slow Internet connections. 

At the same time, just using a streaming http client like in that PR without setting any other kind of timeouts seems dangerous.
@towe75, any suggestions on how to deal with this?

ncode:
Simple change to allow set http timeout via driver config.

Fix #100

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ FEATURES:
 * config: Add `cpu_hard_limit` and `cpu_cfs_period` options [[GH-149](https://github.com/hashicorp/nomad-driver-podman/pull/149)]
 * config: Allow mounting rootfs as read-only. [[GH-133](https://github.com/hashicorp/nomad-driver-podman/pull/133)]
 * config: Allow setting `ulimit` configuration. [[GH-166](https://github.com/hashicorp/nomad-driver-podman/pull/166)]
+* config: Allow setting `image_pull_timeout` and `client_http_timeout ` [[GH-131](https://github.com/hashicorp/nomad-driver-podman/pull/131)]
 * runtime: Add support for host and CSI volumes and using podman tasks as CSI plugins [[GH-169](https://github.com/hashicorp/nomad-driver-podman/pull/169)][[GH-152](https://github.com/hashicorp/nomad-driver-podman/pull/152)]
 
 BUG FIXES:
--- a/README.md
+++ b/README.md
@@ -168,6 +168,14 @@ plugin "nomad-driver-podman" {
 }
 ```
 
+* client_http_timeout (string) Defaults to `60s` default timeout used by http.Client requests
+```
+plugin "nomad-driver-podman" {
+  config {
+    client_http_timeout = "60s"
+  }
+```
+
 ## Task Configuration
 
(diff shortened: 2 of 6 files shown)


jrasell (commented):
Hi @ncode and thanks for taking the time to raise this PR. The code looks good, however, I am wary of the approach as the timeout parameter is set at the plugin level and therefore will impact every HTTP call made.

We had a similar problem on the Docker driver which we [solved by adding a config parameter](https://github.com/hashicorp/nomad/pull/8589) within the job specification driver config itself. This allows overriding this setting per image, rather than on the whole driver. I wonder how you feel about this approach as an addition to the client setting?

The relevant code to this potential approach:
- [api.ImagePull](https://github.com/hashicorp/nomad-driver-podman/blob/057d0c4d444480ae681a1a9f5ea87e374f065088/api/image_pull.go#L14) which accepts a context
- [driver calls ImagePull](https://github.com/hashicorp/nomad-driver-podman/blob/057d0c4d444480ae681a1a9f5ea87e374f065088/driver.go#L694) passing a context

ncode:
> Hi @ncode and thanks for taking the time to raise this PR. The code looks good, however, I am wary of the approach as the timeout parameter is set at the plugin level and therefore will impact every HTTP call made.
> 
> We had a similar problem on the Docker driver which we [solved by adding a config parameter](https://github.com/hashicorp/nomad/pull/8589) within the job specification driver config itself. This allows overriding this setting per image, rather than on the whole driver. I wonder how you feel about this approach as an addition to the client setting?
> 
> The relevant code to this potential approach:
> 
> * [api.ImagePull](https://github.com/hashicorp/nomad-driver-podman/blob/057d0c4d444480ae681a1a9f5ea87e374f065088/api/image_pull.go#L14) which accepts a context
> * [driver calls ImagePull](https://github.com/hashicorp/nomad-driver-podman/blob/057d0c4d444480ae681a1a9f5ea87e374f065088/driver.go#L694) passing a context

Hi @jrasell,

I think the timeout set on the task level sounds better and makes the configuration more granular, so I'm happy to change it. The suggestion from #100 points to plugin level so I followed it :). I will update the PR with the proposed changes.

ncode:
Hi @jrasell,

I've implemented image_pull_timeout, but I also help the client_http_timeout. We need to use both to be able to achieve the timeout from the task. Since the default timeout on the driver for http requests is set to 60s, the context with timeout will drop without the increase. It would be possible to do something similar to https://github.com/hashicorp/nomad/blob/main/api/api.go#L319, but since my previous change helped I decided to keep it.

sycured:
Any ETA about merge/release?

jdoss:
This feature would be great to have. I am trying to launch a ton of Podman containers across a large Nomad cluster and being able to set the timeout would help out a lot.

lgfa29 (approved):
Sorry for getting so long to get this reviewed. I rebased on `main` to get the latest code changes and fix the conflicts.

Thank you so much for the contribution!

PR #137, opened 2021-10-12
config: Privileged containers

towe75:
This PR adds a privileged configuration option, similar to the docker driver.


towe75:
(find my comment about the docs in the nomad repo)

Issue #148, opened 2022-01-05
Running tests locally


optiz0r:
Hi,

Thinking about starting to write some PRs against this project, I was looking to run the tests locally to confirm everything was clean before starting, and that any changes I make still pass tests. I'm having a difficult time getting the tests to run at all, and am seeing lots of failures that don't appear to be happening when running under Github Actions. Are there some undocumented environmental steps required to make tests run?

- `go test` fails with `t.Parallel called multiple times`. Running the test suite with `env CI=1` appears to fix this:

<details>
  <summary>Stack trace</summary>

```
> go test
--- FAIL: TestPodmanDriver_Pull (0.00s)
panic: testing: t.Parallel called multiple times [recovered]
        panic: testing: t.Parallel called multiple times

goroutine 75 [running]:
testing.tRunner.func1.2({0xdec660, 0x10c3090})
        /home/ben/go/src/testing/testing.go:1209 +0x24e
testing.tRunner.func1()
        /home/ben/go/src/testing/testing.go:1212 +0x218
panic({0xdec660, 0x10c3090})
        /home/ben/go/src/runtime/panic.go:1038 +0x215
testing.(*T).Parallel(0xf6d332)
        /home/ben/go/src/testing/testing.go:1077 +0x298
github.com/hashicorp/nomad-driver-podman.startDestroyInspect(_, {{0x1758ff0, 0x4, 0x4}, {{0x0, 0x0}, {0x0, 0x0}}, {0x0, 0x0, ...}, ...}, ...)
        /home/ben/git/nomad-driver-podman/driver_test.go:1973 +0xf6
github.com/hashicorp/nomad-driver-podman.startDestroyInspectImage(0xc000150ea0, {0xf7929f, 0x10}, {0xf728c0, 0x8})
        /home/ben/git/nomad-driver-podman/driver_test.go:1744 +0x165
github.com/hashicorp/nomad-driver-podman.TestPodmanDriver_Pull(0xc000150b60)
        /home/ben/git/nomad-driver-podman/driver_test.go:1738 +0x138
testing.tRunner(0xc000150ea0, 0xfc1918)
        /home/ben/go/src/testing/testing.go:1259 +0x102
created by testing.(*T).Run
        /home/ben/go/src/testing/testing.go:1306 +0x35a
exit status 2
FAIL    github.com/hashicorp/nomad-driver-podman        0.027s
```
</details>

[...]

PR #149, opened 2022-01-06
Add cpu_hard_limit and cpu_cfs_period options

### LINKED ISSUE #147: Add support for CPU hard limits
The docker task driver has support for setting cpu hard limits (added in https://github.com/hashicorp/nomad/pull/3825/files). Podman also supports these via the `--cpu-period=period-in-us` and `--cpu-quota=limit-in-us` options. It would be useful for this driver to support these as well. Currently structs.go is aware of the period and quota options (https://github.com/hashicorp/nomad-driver-podman/blob/dd842857d6f9e928c619801e9f1d31aaf1171709/api/structs.go#L1010), but there is no code to accept a hard limit in the job spec, and apply to the task on creation. I think this would be relatively straightforward to add support for. I may see if I can raise a PR for it myself over the next week or two.

optiz0r:
Nomad's docker driver allows having nomad set cpu resources as hard
limits rather than soft limits via the `cpu_hard_limits` driver config
option. This is useful to restrict how much CPU time a nomad job is
permitted to consume, even if the host is under utilised, or when
predicatibility is more important than speed.

Default behaviour is unchanged. When `cpu_hard_limit` is set to `true`
the podman container is started with the equivalent API options to the
podman cli `--cpu-quota` and `--cpu-period`.

Nomad currently provides a `PercentTicks` value which is used to convert
the cpu resource from the specification into a number of microseconds
of cputime the job is allowed to consume for each `period` interval of time.
Use of this value is discouraged ouside of the docker driver, however the
desired long term replacement is to expose `cpu quota` and `cpu period` as
options in the task `resource` block, and these are not yet currently exposed.
Since this driver emulates the docker interface where possible, I think it's
acceptable to use the `PercentTicks` value in spite of the warning to the contrary.

The `cpu_cfs_period` defaults to 100,000 (microseconds) in both the linux
kernel, and this driver when not specified by the user. User may override
this between 1,000 and 1,000,000. Any value less than 1,000 is silently
increased to 1,000 to prevent
the job failing to run with an `Invalid Operation` error.

The minimum value for the quota is also 1,000. Since this value is
computed by this driver, it is silently increased to 1,000 if the
computed value is lower. For jobs with very small cpu resources in
the job spec, which equate to less than 1% of total available cpu,
the job may consume more cpu time than intended.

Fixes #147

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ FEATURES:
 * config: Map host devices into container. [[GH-41](https://github.com/hashicorp/nomad-driver-podman/pull/41)]
 * config: Stream logs via API, support journald log driver. [[GH-99](https://github.com/hashicorp/nomad-driver-podman/pull/99)]
 * config: Privileged containers.
+* config: Add `cpu_hard_limit` and `cpu_cfs_period` options
 
 BUG FIXES:
 * log: Use error key context to log errors rather than Go err style. [[GH-126](https://github.com/hashicorp/nomad-driver-podman/pull/126)]
--- a/config.go
+++ b/config.go
@@ -47,17 +47,19 @@ var (
 			"username": hclspec.NewAttr("username", "string", false),
 			"password": hclspec.NewAttr("password", "string", false),
 		})),
-		"command":     hclspec.NewAttr("command", "string", false),
-		"cap_add":     hclspec.NewAttr("cap_add", "list(string)", false),
-		"cap_drop":    hclspec.NewAttr("cap_drop", "list(string)", false),
-		"devices":     hclspec.NewAttr("devices", "list(string)", false),
-		"entrypoint":  hclspec.NewAttr("entrypoint", "string", false),
-		"working_dir": hclspec.NewAttr("working_dir", "string", false),
-		"hostname":    hclspec.NewAttr("hostname", "string", false),
-		"image":       hclspec.NewAttr("image", "string", true),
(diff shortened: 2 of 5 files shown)


optiz0r:
@towe75 Could I trouble you to approve the workflow so I can see if the tests pass here? (As noted in #148 I can't get the tests to run locally). Thanks!

optiz0r:
Lint issues fixed and force pushed.

lgfa29:
Thanks for the PR @optiz0r!

I thought that approving it once would do the trick, but GH requires approval at every commit (which makes sense but it's kind of annoying).

To help unblock you, I poked at #148 and was able to run the tests using the provided Vagrantfile in the repo, but I had to run some manual commands. I will post a comment there to keep the discussion separate.

optiz0r:
Perhaps it's not just me - the tests are now failing on GHA with similar results as I was seeing locally.

I think the workflow will still require approval on each push until a change authored in my name is merged in.

towe75:
@optiz0r please rebase to fix the tests.

optiz0r:
@towe75 Any other changes you want to see before merge?

towe75:
@lgfa29 can you have a look, please?

lgfa29 (approved):
Sorry for delay on getting this review. I pushed small refactoring of `setCPUResources` (fewer arguments and less indentation) and also updated the changelog entry to fix a merge conflict and the PR link.

Thanks for the PR @optiz0r 🎉

PR #152, opened 2022-01-23
feat: append nomad task mounts

ttys3:
something like this: https://github.com/hashicorp/waypoint/blob/566c65afb74fcfe64f3f0e50ef7eddbd03d9a3c8/internal/serverinstall/nomad.go#L914

will always fail when using the podman driver, because the driver does not handle this mounts now.

this PR fixup it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ FEATURES:
 * config: Privileged containers.
 * config: Add `cpu_hard_limit` and `cpu_cfs_period` options [[GH-149](https://github.com/hashicorp/nomad-driver-podman/pull/149)]
 * config: Allow mounting rootfs as read-only. [[GH-133](https://github.com/hashicorp/nomad-driver-podman/pull/133)]
+* volume: Add host volumes support for `volume_mount` Stanza [[GH-152](https://github.com/hashicorp/nomad-driver-podman/pull/152)]
 
 BUG FIXES:
 * log: Use error key context to log errors rather than Go err style. [[GH-126](https://github.com/hashicorp/nomad-driver-podman/pull/126)]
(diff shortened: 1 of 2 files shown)


ttys3:
@lgfa29  can you review this ?

lgfa29 (changes requested):
Thanks for the PR @ttys3, and sorry for the delay and getting it reviewed.

I did a quick test using host volumes and it all seems to work 🙂 

Would you mind adding some tests and a `CHANGELOG` entry?

lgfa29:
Ah sorry, I mean like automated testing 😅 

Something like this one, to verify that the configuration is properly set in the driver:
https://github.com/hashicorp/nomad-driver-podman/blob/v0.3.0/driver_test.go#L1225-L1232

You may need to update the `client.hcl` example config file to add a host volume:
https://github.com/hashicorp/nomad-driver-podman/blob/main/examples/nomad/client.hcl

ttys3:
Ok, I'll figure a way to write some test.

lgfa29:
> Ok, I'll figure a way to write some test.

Thanks! I linked to an example, but let me know if you need more help.

lgfa29:
Oh shoot, I forgot about this PR and ended up merging https://github.com/hashicorp/nomad-driver-podman/pull/169, which does the same thing. I'm going to close this one but credit you in the CHANGELOG entry as well.

Sorry for that @ttys3  😞

ttys3:
@lgfa29 sorry for the **huge** delayed unit testing.

I checked the code, almost the same.

but what wondering me is that, I remember what block this PR is **missing unit testing**.

but I did not see any tests code in https://github.com/hashicorp/nomad-driver-podman/pull/169

lgfa29:
Yeah, I wasn't paying enough attention when reviewing #169 😕 

Did you happen to find a way to test this?

PR #141, opened 2021-10-26
Update dependencies

gjpin:
Hello!

I faced the following error when building the driver:
```
go: github.com/hashicorp/go-discover@v0.0.0-20200812215701-c4b85f6ed31f requires
	github.com/tencentcloud/tencentcloud-sdk-go@v3.0.83+incompatible: reading github.com/tencentcloud/tencentcloud-sdk-go/go.mod at revision v3.0.83: unknown revision v3.0.83
```

Updating nomad and go-discover modules solved the issue.

Thanks

References:
https://github.com/hashicorp/nomad/issues/11237#issuecomment-932123501
https://github.com/hashicorp/nomad/pull/11260

PR #155, opened 2022-02-11
Fix broken go-discover dep tree. 

jdoss:
The go-discover mod had some broken deps (https://github.com/hashicorp/go-discover/issues/172) so this PR fixes that issue and updates hclogvet's branch to main. 

Without these changes, anyone building with a clean GOPATH will have a very broken dependency tree.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -17,5 +17,5 @@ deps: ## Install dependencies
 	@echo "==> Installing dependencies..."
 	cd tools && GOBIN=$(CURDIR)/build/bin go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.24.0
 	cd tools && GOBIN=$(CURDIR)/build/bin go get github.com/client9/misspell/cmd/misspell@v0.3.4
-	cd tools && GOBIN=$(CURDIR)/build/bin go get github.com/hashicorp/go-hclog/hclogvet@master 
+	cd tools && GOBIN=$(CURDIR)/build/bin go get github.com/hashicorp/go-hclog/hclogvet@main
 	cd tools && GOBIN=$(CURDIR)/build/bin go get gotest.tools/gotestsum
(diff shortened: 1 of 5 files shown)


lgfa29:
Thanks for the PR @jdoss!

Kind of weird that `go-discover` is being pulled, but I guess it's because of `github.com/hashicorp/nomad`. I wasn't able to reproduce the broken build, so would you mind testing upgrading `github.com/hashicorp/nomad` instead and see if it pulls the correct `go-discover` version?

lgfa29:
Just noticed https://github.com/hashicorp/nomad-driver-podman/pull/141, maybe updating just Nomad is not enough?

Still, I would be curious to see if it works 🙂

ttys3:
I can't reproduce the issue.

jdoss:
@lgfa29 I tried just upgrading Nomad but it wasn't enough. This was the only way I could get past the deps to get this plugin to build.

@ttys3 are you doing with a totally empty GOPATH? If your current GOPATH has the old module it will build just fine.

lgfa29:
Ah ok, thanks for trying 🙂 

I wasn't able to reproduce it either, but it was also reported by another user (#141), so there may be something weird going-on.

This is what I tried to get a clean environment:

```console-shell
$ docker run --rm -it golang:1.14 /bin/bash
$ git clone https://github.com/hashicorp/nomad-driver-podman.git src/hashicorp/nomad-driver-podman
$ cd src/hashicorp/nomad-driver-podman/
$ ./build.sh
```

I tried Go 1.17 as well and it built fine 🤔

jdoss:
Do you folks still want this PR? I can close it if not.

lgfa29:
Thanks for the ping @jdoss, but I haven't been able to reproduce the issue, even in a clean environment.

@towe75 have seen this issue before?

towe75:
@jdoss thank you for the PR

@lgfa29 i can not reproduce it in my personal dev env and also not in our (private) nightly builds. We're building on ephemeral nodes with golang 1.17 there. Also the GH Action Runner has no problem. 

My 2 Cents: merging it will not hurt us and it might fix a subtle problem for some users.

lgfa29 (approved):
Sounds good to me @towe75.

Thanks for the PR @jdoss, sorry it took this long to get it merged 😅

PR #157, opened 2022-02-25
Add syntax highlighting for `hcl` and `sh` code in README.md

EllipticCrab:
Seems like the README.md file had some commands with the `sh` syntax highlighting, but not all. Also, the HCL examples were missing highlighting. This PR adds them to make the rendered README look a bit nicer 🙂 

Link to rendered file: https://github.com/hashicorp/nomad-driver-podman/blob/4486e3bae6da45a330c8c7ed7f3fd10b74fd5ffb/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ this plugin to Nomad!
 
 Here is a simple redis "hello world" Example:
 
-```
+```hcl
 job "redis" {
   datacenters = ["dc1"]
   type        = "service"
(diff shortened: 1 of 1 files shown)


jrasell (approved):
LGTM, thanks @EllipticCrab!

lgfa29 (approved):
Thanks @EllipticCrab!

I pushed a commit to use `console` instead of `sh` in some outputs that are about running commands.

EllipticCrab:
Ah, TIL about `console`. Makes sense, thanks!

PR #159, opened 2022-03-03
Prevent concurrent image pulls of same imageRef

towe75:
podman does not establish a lock when it pulls an image. 

There is some movement towards a more efficient implementation via https://github.com/containers/podman/issues/1911
Other people workaround this problem by having a lock in another layer like e.g. https://github.com/cri-o/cri-o/pull/3409

This PR adds such a mechanism to nomad-driver-podman. It prevents to pull the same image on the same host concurrently. A good, manual, check is to run a job with e.g. 5x redis and to ensure that the image is locally not available. First allocation will pull the image, 4 others will wait until it is ready. Without this PR we would try to download it 5 times parallel, wasting time and bandwith.

There is no unit test because i could not figure out a good way to code it. Maybe i can get a recommandation?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+IMPROVEMENTS:
+
+* runtime: Prevent concurrent image pulls of same imageRef [[GH-159](https://github.com/hashicorp/nomad-driver-podman/pull/159)]
+
 ## 0.4.0 (July 14, 2022)
 
 FEATURES:
--- a/driver.go
+++ b/driver.go
@@ -30,6 +30,7 @@ import (
 	"github.com/hashicorp/nomad/plugins/shared/hclspec"
 	pstructs "github.com/hashicorp/nomad/plugins/shared/structs"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
+	"golang.org/x/sync/singleflight"
 )
 
 const (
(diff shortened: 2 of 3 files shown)


lgfa29:
Hi @towe75 👋 

Sorry for the delay in getting this reviewed 😅 

The changes look good, but I think we can get the same behaviour in a simpler way by using the [`singleflight`](https://pkg.go.dev/golang.org/x/sync/singleflight) package. Here's how that would look like: 
https://github.com/hashicorp/nomad-driver-podman/compare/f_lock_pull_singleflight?expand=1

The main difference with that approach is that we can't really tell if there's another tasking downloading the image, the package will just block when `Do` is called. But I think that's fine, and even a little desirable as it hides the inner workings from the user?

From their perspective, they would just see this for all tasks:
image

So it doesn't really matter which task actually downloaded the image.

With regards to testing, yeah, I can't think of an easy way to test this 🤔 

One idea was to somehow mock and wrap the `ImagePull` method from the `api` package in a way that we can count how many times it's called, then we assert that it was only called once. But I think that would require quite a bit of changes to make an interface etc.

towe75:
@lgfa29 sorry, the "compare" link does not work for me? I will have a look at the `singleflight` package, thank you.

towe75:
It uses now the `singleflight` package. Code is much shorter now but it also lacks one driver detail event compared to my previous version. This event was just cosmetic, so i am fine with not having it.

@lgfa29  please review again.

lgfa29 (approved):
Agree, that event was more of an implementation detail that was not relevant to users. 

Just missing a CHANGELOG entry, but looks good and much simpler 🙂

PR #150, opened 2022-01-07
update Vagrantfile and VM test instructions

lgfa29:
The OpenSuse registry CA certificates need to be updated before adding it to `apt`.

I also added the commands that need to be executed for running tests following what's set in the GitHub Action.

PR #164, opened 2022-04-26
crt: onboard nomad-driver-podman to CRT

shoenig:
Migrate build pipeline to the new Common Release Tooling process. During this process we also had to:

- Updates the Nomad dependency to 1.3.1 to fix some build issues with `github.com/opencontainers/runc`.
- Rename and update the test workflow so it also runs the linters
- Update `go.mod` file format to 1.17
- Disable OSV security scan due to some unrelated security issues (I think these will go away once update some dependencies)
- Add new make targets to standardize them across projects


lgfa29:
Test errors are caused by an inconsistent mirror. I have a fix for it here:
https://github.com/hashicorp/nomad-driver-podman/pull/150

lgfa29:
Test errors are caused by an inconsistent mirror. I have a fix for it here:
https://github.com/hashicorp/nomad-driver-podman/pull/150

lgfa29 (approved):

PR #177, opened 2022-06-29
Replaced generic error messages

unexist:
I am struggling to run this driver on macOS and one of the real problems is every error message is basically the same generic "unknown error". 

So this PR just adds the command to the error to make it easier to understand what is actually happening.

### DIFF
--- a/api/container_create.go
+++ b/api/container_create.go
@@ -28,7 +28,7 @@ func (c *API) ContainerCreate(ctx context.Context, create SpecGenerator) (Contai
 
 	if res.StatusCode != http.StatusCreated {
 		body, _ := ioutil.ReadAll(res.Body)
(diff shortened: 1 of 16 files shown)


unexist:
Hey, 

glad I could help and thanks for the work done so far. :)

Well, running Nomad with Podman on macOS is quite a challenge. One of my main learnings so far is you have to reverse mount your home or rather the allocation path of Nomad into the Podman VM, otherwise the allocations fail with a really cryptic error message. (podman machine init -v $HOME:$HOME)

I'll probably open up another PR with a bit of documentation, once I understand my more serious problem:

When I start a job/container via this driver Podman itself becomes unresponsive and I cannot interact with it at all anymore e.g. via its unix socket. Any hints how to debug this?

Issue #176, opened 2022-06-24
Ship podman driver in official nomad distribution


svenstaro:
Currently nomad is distributed separately but I think with podman receiving so much support and attention lately, it would be prudent to consider shipping it in the official nomad distribution to make it less of a hassle to install and to signal official support. It's already listed in the docs along with the other official drivers so I think it would make sense.

lgfa29:
Hi @svenstaro 👋 

Thanks for opening this issue. While we understand it would make it easier to use the plugin the Podman API still changes fairly often and we still need to catch-up. Having the plugin embedded would also tie the release cycle of the projects, which is something we try to avoid until plugins mature a little more.

I will leave this open as it will be good to collect more feedback, but I just want to make sure to set the right expectations that we don't currently have a plan for this.

PR #181, opened 2022-07-17
enable building rpm and deb through github actions

flowpoint:
Hi there,
it seemed like there aren't yet packages for the podman driver.
it would be a nice to have in your yum and deb repos, like https://rpm.releases.hashicorp.com/fedora/hashicorp.repo](https://rpm.releases.hashicorp.com/fedora/hashicorp.repo)

i hope this pr is a start.
this might also be a solution for  #176

i tried to match your release setup from [the nomad repo](https://github.com/hashicorp/nomad).
however, [hashicorp/actions-packaging-linux](https://github.com/hashicorp/actions-packaging-linux) expects there to be a binary and installs it only **/usr/bin**.
i lifted this requirement in my temporary fork, see: [flowpoint/actions-packaging-linux](https://github.com/flowpoint/actions-packaging-linux) and opened [https://github.com/hashicorp/actions-packaging-linux/pull/7](https://github.com/hashicorp/actions-packaging-linux/pull/7)

comments appreciated
greetings

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,3 +88,35 @@ jobs:
         with:
           name: ${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
           path: ${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
+
+
+      - name: Package
+        uses: flowpoint/actions-packaging-linux@v1.6
+        with:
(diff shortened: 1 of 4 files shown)


lgfa29:
Hi @flowpoint 👋 

Thanks for the PR! We currently don't package any of our plugins, but it's an interesting suggestion. I think it would require some sort of standardization and coordination to make sure all plugins are installed the same way. I will discuss it with the rest of the team.

pkazmierczak:
hi @flowpoint, sorry for getting back to you so late. We recently connected the podman driver repository to our internal package building pipeline, and the plugin packages are now available in our apt and yum registries.

PR #186, opened 2022-07-26
Use ping api instead of system info for fingerprinting

towe75:
## Overview

All nomad drivers must implement a fingerprint mechanism. Back at the day i went with the system info api call and it worked well.  

But it turned out that this is a rather heavy invocation, going even down to the package manager level (dpkg, rpm) to gather informations. The driver builds a fingerprint each 30 seconds and thus it wastes quite some cpu cycles and has an impact on the overall machine load.

## Some numbers

A simple `perf stat` on my machine shows the difference between the two podman cli equivalents:

```
 Performance counter stats for 'podman info':

            240.46 msec task-clock                #    0.721 CPUs utilized
             1,481      context-switches          #    6.159 K/sec
               141      cpu-migrations            #  586.383 /sec
            18,851      page-faults               #   78.397 K/sec

       0.333681791 seconds time elapsed

       0.098392000 seconds user
       0.172141000 seconds sys
```

```
 Performance counter stats for 'podman version':

             60.49 msec task-clock                #    0.426 CPUs utilized
               720      context-switches          #   11.903 K/sec
                74      cpu-migrations            #    1.223 K/sec
             4,449      page-faults               #   73.549 K/sec

       0.141960043 seconds time elapsed

       0.036505000 seconds user
       0.047359000 seconds sys
```

## What the PR does

This PR changes the mechanism to use the _ping api call which is not so resource intensive. System info is now usually  only called once. 

## Conclusion

You will not see a big performance different on a crowded host but it certainly makes the situation better on a idle machine.

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -237,35 +237,32 @@ func (d *Driver) buildFingerprint() *drivers.Fingerprint {
 			Health:            drivers.HealthStateUndetected,
 			HealthDescription: "disabled",
 		}
-	} else {
+	}
 
-		// it's reachable
-
-		// do we already know details about podman or is the version different?
-		if d.systemInfo.Version.APIVersion != version {
-			// no? then fetch and cache it
-			// try to connect and get version info
-			info, err := d.podman.SystemInfo(d.ctx)
-			if err != nil {
-				d.logger.Error("Could not get podman info", "error", err)
-				return &drivers.Fingerprint{
-					Attributes:        attrs,
-					Health:            drivers.HealthStateUndetected,
-					HealthDescription: "disabled",
-				}
-			} else {
-				// keep first received systemInfo in driver struct
-				// it is used to toggle cgroup v1/v2, rootless/rootful behavior
-				d.systemInfo = info
-				d.cgroupV2 = info.Host.CGroupsVersion == "v2"
+	// do we already know details about podman or is the version different?
+	if d.systemInfo.Version.APIVersion != version {
+		// no? then fetch and cache it
+		// try to connect and get version info
+		info, err := d.podman.SystemInfo(d.ctx)
+		if err != nil {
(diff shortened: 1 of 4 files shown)


driver.go
```diff
@@ -223,38 +223,54 @@ func (d *Driver) handleFingerprint(ctx context.Context, ch chan<- *drivers.Finge
...
+		// not reachable?
+		// deactivate driver, forget podman details
+		d.systemInfo = api.Info{}
+		d.logger.Error("Could not get podman version", "error", err)
+		return &drivers.Fingerprint{
+			Attributes:        attrs,
+			Health:            drivers.HealthStateUndetected,
+			HealthDescription: "disabled",
+		}
 	} else {
```
lgfa29, line 240: Minot nit-picking: I think this `else` is unnecessary since the `if` body will `return`? This would make the rest of the code less indented 🙂

driver.go
```diff
@@ -223,38 +223,54 @@ func (d *Driver) handleFingerprint(ctx context.Context, ch chan<- *drivers.Finge
...
+			// try to connect and get version info
+			info, err := d.podman.SystemInfo(d.ctx)
+			if err != nil {
+				d.logger.Error("Could not get podman info", "error", err)
+				return &drivers.Fingerprint{
+					Attributes:        attrs,
+					Health:            drivers.HealthStateUndetected,
+					HealthDescription: "disabled",
+				}
+			} else {
```
lgfa29, line 256: Same for this `else` as well.

lgfa29 (approved):
Minor nit-picking, but feel free to ignore them 🙂

driver.go
towe75: Reasonable. Refactored code multiple times and did not see this simplification, thx.

PR #183, opened 2022-07-21
fix container stats endpoint response handling

dermetfan:
Podman 4.1.1 changed the `stats` endpoint's HTTP status for stopped containers from 404 to 200.

Fixes #182
Closes #187

PR #187, opened 2022-08-07
fix: podman v4.1.1 or newer version stats api

ttys3:
podman v4.1.1 or newer version will return 200 code and empty body if the container is exited

which will cause nomad nerver got notified even if the container exited with 0

when using lifecycle, the leader task will wait forerver because nomad think the sidecar=false container is still running
due to podman did not report the state to nomad.

for old podman version, it will return `http.StatusConflict` 409, so this fallback fixup is OK to the old version.

Refs: https://github.com/containers/podman/issues/14498

https://github.com/containers/podman/issues/15218


zandeez:
duplicate of #183

Issue #184, opened 2022-07-22
Setting selinuxlabel="z" in the client driver config breaks CSI hostpath plugin job


jdoss:
I was trying to get the csi-hostpath plugin working on Fedora CoreOS which has SELinux set to enforcing by default. I kept getting this error:

```
rpc error: code = Unknown desc = failed to start task, could not start container: cannot start container, status code: 500: {"cause":"SELinux relabeling of /dev is not allowed","message":"SELinux relabeling of /dev is not allowed","response":500}
```

Removing `selinuxlabel = "z"` from the plugin section of the client config allows the job to start without this error. 

```
plugin "nomad-driver-podman" {
          config {
            volumes {
              enabled      = true
              selinuxlabel = "z"
            }
          }
        }
```

Now I have to explicitly set on volumes 
```
config {
        image = "docker.io/fluent/fluentd:latest"
        volumes = [
          "local/fluent.conf:/fluentd/etc/fluent.conf:z"
        ]
      }
```

which isn't a huge deal, but we should document this in the README (I can PR that if that's the best course of action) or not apply the automatic selinuxlabel to CSI plugin jobs (I am not sure if that is even possible).

jdoss:
This might fix things here too? https://github.com/hashicorp/nomad-driver-podman/pull/66

jdoss:
@tgross can #66 get merged? I just verified that it fixes this issue.

tgross:
It never got reviewed for some reason and currently has a merge conflict. This week the team has their plate full with the Nomad 1.4.0 release but I'll try to nudge folks about taking a look soon.

PR #66, opened 2020-09-12
Moves SELinux label assignment to affect only implicit mounts

fuero:
Fixes #65 by moving the SELinux label assignment above the handling of user-supplied volumes.

Issue #65, opened 2020-09-08
SELinux labeled volumes on shared filesystem


fuero:
OS version: CentOS Linux release 8.2.2004 (Core)
nomad-driver-podman version: git commit 49d9894
nomad version: 0.12.3
podman version: 1.6.4, shipped by CentOS 8
gluster version: 7.7

I got 2 nomad nodes on CentOS 8 set up to mount a gluster volume:

```
gluster.service.local-consul:/gv0 /shared 	glusterfs context=system_u:object_r:container_file_t:s0,selinux 0 0
```
`ls` shows the correct context is applied:
```
# ls /mnt/bar
drwxr-xr-x.  1 root root system_u:object_r:container_file_t:s0    4.0K Sep  7 16:15 bar
```
Running chcon on that filesystem produces an error:

```
# chcon -u system_u -r system_r -t container_file_t -l s0 /mnt/bar
chcon: failed to change context of '/mnt/bar' to ‘system_u:system_r:container_file_t:s0’: Operation not supported
```

Setting `selinuxlabel = "z"` won't let me start *any* container that has volumes on that share, as podman forces relabeling, triggering this error:

```
rpc error: code = Unknown desc = failed to start task, could not start container: io.podman.ErrorOccurred(Reason: relabel failed "/mnt/foo": operation not supported)
```

According to the nomad-driver-podman's docs, `selinuxlabel` is supposed to apply *only* to the local, secrets and alloc mounts:

```
Allows the operator to set a SELinux label to the allocation and task local bind-mounts to containers
```

If the job in question is placed on a filesystem that supports SELinux labels it works fine.

Expected results:

podman not being forced to relabel on a volume that's properly prepared.

lgfa29:
Fixed in #196.

Issue #185, opened 2022-07-22
Cannot use volume and volume_mount when SELinux is set to enforcing. 


jdoss:
It seems that I can't mount any kind of directory with the CSI hostpath plugin using `volume` and `volume_mount` directives. 

```
group "victoriametrics" {
    count = 1

    restart {
      attempts = 10
      interval = "5m"
      delay = "25s"
      mode = "delay"
    }

    network {
      port "http" {
        to = 8428
      }
    }

    service {
      name = "victoriametrics"
      tags = ["metrics", "victoriametrics"]
      meta {
        nomad_alloc_id = "${NOMAD_SHORT_ALLOC_ID}"
        }
      provider = "consul"
      port = "http"
    }

    service {
      name = "influxdb"
      tags = ["metrics", "influxdb"]
      meta {
        nomad_alloc_id = "${NOMAD_SHORT_ALLOC_ID}"
        }
      provider = "consul"
      port = "http"
    }

    volume "victoriametrics" {
      type      = "csi"
      source    = "victoriametrics"
      read_only = false
      attachment_mode = "file-system"
      access_mode     = "single-node-writer"
    }

    task "http" {
      driver = "podman"

      volume_mount {
        volume      = "victoriametrics"
        destination = "/victoria-metrics-data"
        read_only   = false
      }

      config {
        image = "docker.io/victoriametrics/victoria-metrics:v1.79.0"
        ports = ["http"]

        # volumes = [
        #   "/opt/nomad/storage/victoria-metrics-data:/victoria-metrics-data:z"
        # ]
      }

      resources {
        cpu    = 600 # MHz
        memory = 768 # MB
      }
    }

With the above config the allocation fails over and over again. Disabling SELinux client node wide (something I'd like to avoid) or manually adding `container_file_t` to `/opt/nomad/data/` (which is effectively `:z` on a volume mount) to my nomad data directory works around the issue. 

```
# chcon system_u:object_r:container_file_t:s0 /opt/nomad/data/ -R
```

I believe this is related to the bigger conversation in https://github.com/hashicorp/nomad/issues/9123 but maybe https://github.com/hashicorp/nomad-driver-podman/pull/139 could get merged so I could use `security_opt = ["label=disable"]` as a stop gap to disable SELinux on my container workloads that need persistent storage.

Ideally, it would be cool to just add `selinuxlabel = "z"` to volume_mount blocks so we can handle SELinux labeling:

```
volume_mount {
[...]

PR #196, opened 2022-10-20
Fix logic on volume SELinux labels

jdoss:
This will fix https://github.com/hashicorp/nomad-driver-podman/issues/184 by not applying SELinux contexts to privileged containers (CSI plugins run as privileged containers) and it will also allow for CSI plugins to correctly label SELinux contexts when mounting storage into a container when SELinux is set to enforcing.

This should close https://github.com/hashicorp/nomad-driver-podman/pull/66 as this PR would prevent SELinux contexts from being applied to CSI mounts and they would be unusable on hosts that have SELinux set to enforcing.

I have tested this with the https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver and I am able to read and write to volume mounts from this CSI plugin as expected:

```
# ls -lahZ /opt/nomad/data/client/csi/node/gcp-pd-csi/per-alloc/ec76605e-24c6-8fe4-1a84-a6d8e6a9a6ff/nginx-selinux-test/rw-file-system-single-node-writer/                          
total 20K
drwxr-xr-x. 3 root root system_u:object_r:container_file_t:s0 4.0K Oct 20 19:04 .
drwx------. 3 root root system_u:object_r:var_t:s0              47 Oct 20 19:24 ..
-rw-r--r--. 1 root root system_u:object_r:container_file_t:s0    0 Oct 20 19:04 bye.txt
-rw-r--r--. 1 root root system_u:object_r:container_file_t:s0    0 Oct 20 14:11 hello.txt
drwx------. 2 root root system_u:object_r:container_file_t:s0  16K Oct 20 11:31 lost+found
```

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -1179,7 +1179,7 @@ func (d *Driver) containerMounts(task *drivers.TaskConfig, driverConfig *TaskCon
 		binds = append(binds, bind)
 	}
 
-	if selinuxLabel := d.config.Volumes.SelinuxLabel; selinuxLabel != "" {
+	if selinuxLabel := d.config.Volumes.SelinuxLabel; selinuxLabel != "" && !driverConfig.Privileged {
 		// Apply SELinux Label to each volume
 		for i := range binds {
 			binds[i].Options = append(binds[i].Options, selinuxLabel)


jdoss:
Fixes https://github.com/hashicorp/nomad-driver-podman/issues/65

jdoss:
I will note that I saw this error when trying to start a CSI controller:

```
Oct 21, '22 13:39:14 -0500 	Driver Failure 	rpc error: code = Unknown desc = failed to start task, could not start container: cannot start container, status code: 500: {"cause":"SELinux relabeling of /dev is not allowed","message":"SELinux relabeling of /dev is not allowed","response":500}
```

which I thought was odd because the [csi disk documentation](https://learn.hashicorp.com/tutorials/nomad/stateful-workloads-csi-volumes#deploy-the-ebs-plugin) says that only the nodes need be running as privileged, but it seems that both `node` and `controller` plugins are mounting `/dev/` as `rw` when being registered as CSI plugins. 

Running both the controller and the node tasks as `privileged = true` allows everything to work correctly with SELinux enabled with this driver.

@tgross should CSI controllers need `rw` access to `/dev`? If not, that feels like a Nomad specific bug and I can open an issue on the Nomad repo. Since controllers are not doing the mounting, the shouldn't need access to `/dev` right?

Here is my GCP PD job file:

```
job "csi-gcp-pd-plugin" {
  datacenters = ["us-central1"]
  type = "system"

  group "controller" {
    task "gcp-controller-plugin" {
      driver = "podman"
      config {
        image = "k8s.gcr.io/cloud-provider-gcp/gcp-compute-persistent-disk-csi-driver:v1.7.2"
        args = [
          "--endpoint=unix:///csi/csi.sock",
          "--v=6",
          "--logtostderr",
          "--run-node-service=false"
        ]
        privileged = true
      }
      env {
        GOOGLE_APPLICATION_CREDENTIALS = "/secrets/gcp-pd-csi.json"
      }
      template {
        destination = "secrets/gcp-pd-csi.json"
        data        = <<-EOF
        {{ "my-base64-encoded-gcp-service-account-json" | base64Decode }}
        EOF
      }

      csi_plugin {
        id        = "gcp-pd-csi"
        type      = "controller"
        mount_dir = "/csi"
      }
      resources {
        memory = 256
      }
    }
  }
  group "node" {
    task "gcp-node-plugin" {
      driver = "podman"
      config {
        image = "k8s.gcr.io/cloud-provider-gcp/gcp-compute-persistent-disk-csi-driver:v1.7.2"

        args = [
          "--endpoint=unix:///csi/csi.sock",
          "--v=6",
          "--logtostderr",
          "--run-controller-service=false"
        ]
        privileged = true
      }
      env {
          GOOGLE_APPLICATION_CREDENTIALS = "/secrets/gcp-pd-csi.json"
      }
      template {
          destination = "secrets/gcp-pd-csi.json"
          data        = <<-EOF
          {{ "my-base64-encoded-gcp-service-account-json" | base64Decode }}
          EOF
      }
      csi_plugin {
        id        = "gcp-pd-csi"
        type      = "node"
        mount_dir = "/csi"
      }
      resources {
        memory = 256
      }

    }

  }
}
```

tgross:
> @tgross should CSI controllers need rw access to /dev? If not, that feels like a Nomad specific bug and I can open an issue on the Nomad repo. Since controllers are not doing the mounting, they shouldn't need access to /dev right?

The specification is ambiguous on what "controller publish" actually _means_ and leaves it up to the plugin, so yeah it's entirely possible we've hit a controller plugin that requires it in the case of selinux relabeling.

lgfa29 (changes requested):
Thanks for the PR and all the context provided in the linked issues @jdoss!

I'm not well versed in SELinux, so apologies in advance for my ignorance and anything wrong I say 😅 

From what I understood the root of the problem is that the hots path of volumes in file systems that support SELinux must be mounted with the `z` label so that the runtime can properly set their policy to allow the container to access it.

By using a driver-level config of `volumes.selinux_labels` we don't have enough fine grain control to disable this when necessary (such as when running CSI plugins, or using plugins that don't support SELinux). I did a quick research on how Kubernetes handle this, and [they seem to hardcode which plugins support and which ones don't](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/1710-selinux-relabeling/README.md#selinux-support-in-volumes), which doesn't seem like a good approach for us.

Using the `privileged` flag to check doesn't also seem ideal, as it's just loosely correlated with the problem.

Another thing I noticed is that this logic is pretty close to the [Docker driver](https://github.com/hashicorp/nomad/blob/v1.4.2/drivers/docker/driver.go#L702-L707), and so I think there's a broader issue to be solved here.

Indeed, https://github.com/hashicorp/nomad/issues/9123 points to a similar problem as you described in https://github.com/hashicorp/nomad-driver-podman/issues/185 and https://github.com/hashicorp/nomad/pull/7094 has a similar changed as proposed in https://github.com/hashicorp/nomad-driver-podman/pull/66.

You made an interesting suggestion in #185:

> Ideally, it would be cool to just add `selinuxlabel = "z"` to `volume_mount` blocks so we can handle SELinux labeling [...] on a per mount basis.

This could be one way forward. We have [`volume.mount_options`](https://developer.hashicorp.com/nomad/docs/job-specification/volume#mount_options) so I was wondering if we could add a new field to store task driver specific flags, or something like that.

But coming back to the issue in hand, does the procedure you described in https://github.com/hashicorp/nomad-driver-podman/issues/184 work in general? Maybe we can document it for now until we come up with a better solution.

jdoss:
Hey @lgfa29 I agree that there is a better path forward w/r/t SELinux and volume mounts in general, but it will be more involved of a fix.

The issue is that some jobs need to be run as privileged containers and using `selinuxlabel = "z"` setting in the plugin section applies this context to every mount. This is going to be problematic because the `--privileged` flag adds mounts such as `/dev` that cannot be relabeled. Also, privileged containers don't need this label added because they run without the constraints that SELinux applies. 

>Executing container engines with the --privileged flag tells the engine to launch the container process without any further "security" lockdown.

https://www.redhat.com/sysadmin/privileged-flag-container-engines

I have been running this patch in production for weeks now and it works as expected. I hope we can get this merged so it unblocks the usage of CSI plugins with this driver until a better solution is implemented which would give us better control on how `selinuxlabel` is applied to volume mounts. I do think that if a container is run with `--privileged` it should be excluded anyways because it doesn't need anything from `selinuxlabel`. It already has it all the privileges it needs from the `--privileged` flag.

jdoss:
I can push a PR up that should address the issue in the Docker driver, but I don't run Docker on my systems anymore so testing it would be an issue. I am also pretty backlogged right now so if you beat me to getting it patched that works too.

We can close this issue https://github.com/hashicorp/nomad-driver-podman/issues/65 too since its was the original issue and its PR was superseded by this one.

lgfa29:
No worries, I will see if I can repro the problem in Docker and submit a fix, should be straightforward with all the information you provided us 👍

PR #139, opened 2021-10-25
Add support for container creation's selinux_opts attribute

gjpin:
Solves #135 

Hi!

This PR adds support for the selinux_opts attribute, as per https://docs.podman.io/en/latest/_static/api.html#operation/ContainerCreateLibpod .

I've also updated the README file and the tests to support the new attribute.

Thank you!

PR #199, opened 2022-11-14
ci: skip SELinux tests since they are not supported in GHA

lgfa29:
I merged #139 without checking the actual test result 🤦 

@jdoss did a great job [investigating how the SELinux project runs their CI](https://github.com/hashicorp/nomad-driver-podman/pull/139#issuecomment-1192929834), but running a Vagrant machine in our case seemed a bit much so I opted to just skipping these tests in CI for now.


jrasell (approved):
LGTM!

PR #204, opened 2022-11-22
Set mount propagation from TaskConfig

### LINKED ISSUE #192: Podman does not support the mount propagation needed by CSI plugins
In #169 @tgross added the ability for the Podman driver to bind mounts from a TaskConfig. While this didn't allow the addition of a mount block using the external API (that's left to do in #142), it did allow Podman to set up the bind mounts requested by the [CSI Plugin Prestart Hook](https://github.com/hashicorp/nomad/blob/main/client/allocrunner/taskrunner/plugin_supervisor_hook.go#L143-L211), enabling Podman driver CSI support.

As we were experimenting with this functionality for the BeeGFS CSI driver, we noticed that our driver container seemed unable to see volumes it had already mounted. In NodeUnpublishVolume and NodeUnstageVolume, this manifested itself as a bunch of errors wherein Nomad would tell the driver to unstage but the driver insisted there was no mount to operate on. An inspection of the driver container shows the following:

```
 "Binds": [
                "/opt/nomad/alloc/edb56316-6229-5b3a-e5eb-6053a698d2b7/alloc:/alloc:rw,rprivate,rbind",
                "/opt/nomad/alloc/edb56316-6229-5b3a-e5eb-6053a698d2b7/node/local:/local:rw,rprivate,rbind",
                "/opt/nomad/alloc/edb56316-6229-5b3a-e5eb-6053a698d2b7/node/secrets:/secrets:rw,rprivate,noexec,rbind",
                "/:/host:ro,rshared,rbind",
                "/opt/nomad/client/csi/plugins/edb56316-6229-5b3a-e5eb-6053a698d2b7:/csi:rw,rprivate,rbind",
                "/opt/nomad/client/csi/node/beegfs-csi-plugin:/opt/nomad/client/csi/node/beegfs-csi-plugin:rw,rprivate,rbind",
                "/dev:/dev:rw,rprivate,nosuid,rbind"
            ],
```

There's a lot there, but the important bit is that `/opt/nomad/client/csi/node/beegfs-csi-plugin` is bind mounted to `/opt/nomad/client/csi/node/beegfs-csi-plugin` as rprivate (this would be `/opt/nomad/client/csi/node/beegfs-csi-plugin` to `/local/csi` for most plugins, but this is just a BeeGFS CSI driver implementation detail). The [CSI Plugin Prestart Hook](https://github.com/hashicorp/nomad/blob/54474a9534ca8f105ca5c35ec2f11a3f7602c485/client/allocrunner/taskrunner/plugin_supervisor_hook.go#L169-L173) wants this to be a bidirectional mount (rshared), but the changes in #169 don't take PropagationMode into account. Unless it the bind has rshare, it's not clear how most CSI drivers would be able to propagate the mounts they create inside their containers to Nomad tasks.

I think a fix for this would be as easy as adding a conversion like the [Docker driver already does](https://github.com/hashicorp/nomad/blob/9504a4537b996820d942d637ac8c65112e04d88d/drivers/docker/driver.go#L712-L719), but the Docker driver has some [extra logic in it](https://github.com/hashicorp/nomad/blob/9504a4537b996820d942d637ac8c65112e04d88d/drivers/docker/driver.go#L1062-L1066) that gives me a bit of pause (can the Podman driver run on Windows?).

I can probably find some time to make a quick change and do some sanity checking, but am I thinking about this issue correctly?

ejweber:
Fixes #192.

See the issue for full details. As a quick summary, CSI drivers running as containers need correct mount propagation (between their internal mount namespace and the host's) in order to function correctly. Until Nomad v1.4.3, we did not have access to mount propagation information in our TaskConfig. Now we do, so this PR simply translates that received information into the bind mounts Podman sets up.


tgross (approved):
LGTM! 

Would you mind adding an extra to the `CHANGELOG.md` file as well?

ejweber:
> Would you mind adding an extra line to the `CHANGELOG.md` file as well?

Just did this with an amended commit.

PR #205, opened 2022-12-01
support for rootless with cgroups_v1 by ignoring memory and cpu settings

rikislaw:
When we want to start podman container in rootless mode and without cgroups_v2, then we have error message like below:
Error: invalid configuration, cannot specify resource limits without cgroups v2 .....

Error message is different depending on used runc version. The same happens when we are using Nomad to start job with podman driver.  This happens at least on RHEL7 & RHEL8 (without enabled cgroups v2).

Proposed change ignores setting cpu and memory limits when we have rootless mode with cgroups v1.

BR,
Piotr Richter


mikenomitch:
Hey @rikislaw,

Firstly, thanks for the PR and apologies for the delay in reviewing it. We had a discussion on this internally last week, but we forgot to follow up with an answer. I understand that there are some other tweaks to the driver that you’re interested in, hopefully we can have a quicker turnaround on those.

We are worried about the approach that this takes though. The concern is that Nomad won’t be able to make guarantees about resource utilization for any task if a single task driver ignores resource reservations. So this could have negative spillover effects to other jobs.

I see in the description above that you’re trying to go without cgroups_v2. Is that a hard requirement on your end? The use of cgroups v2 is effectively a requirement for managing rootless containers with resource isolation (https://rootlesscontaine.rs/getting-started/common/cgroup2/). That would still need a small change from Nomad to work, but it should be doable on our end.

- Mike

rikislaw:
Hi Mike,

To be honest I'm also worried, that I need to use it;) But currently we are using RHEL7.9, where it is not possible (due to old kernel) to enable cgroups_v2. We are also obligated to use rootless mode. This means, that with current podman driver 4.1 we are not able to start any container, because Nomad is always setting limits and podman (runc) responds with an error (Error: invalid configuration, cannot specify resource limits without cgroups v2 .....).

In PR I'm adding condition for cgroup_v1 and rootless, where setting limits is not possible at all, so it doesn't break anything. I understand, that it is not ideal from Nomad perspective, because Nomad thinks, that there are some limits. Maybe it would be more elegant and clear to add for podman driver option ignore_resources_limits (default: false). Then everyone by using this flag would be fully aware what he is doing.

We need this or similar solution before moving to RHEL8, where it will be possible to enable cgroups_v2, but it will take some time and till then we would like to use RHEL7 with Nomad and Podman.

If this is not going to be approved, then we will be forced to develop podman-driver by ourselves.

BR,
Piotr

tgross:
Hi @rikislaw! One of the concerns we have with this is that such an option would break other compliance controls that users have, like Quotas or Sentinel policies. For example, you could schedule your job with 10MB of memory to get around quotas and then let the job use up GBs of memory anyways. I'm pretty sure under such a model everything would end up in the same cgroup slice as well, which means things like oomd would potentially kill _the entire client and all podman tasks_ in the case of high memory use. And that's a DoS concern.

You say you're stuck on an older version of RHEL, but is it possible for you to use the cgroup v2 boot arg with that version? That would let you have safe rootless podman, keep the version of RHEL, and avoid having `unsafe_dont_use_no_resource_controls` (:grinning:) config flags added to the driver.

rikislaw:
Hi @tgross ,

Unfortunately RHEL7 is not supporting Cgroups_v2. It has too old kernel for this. So if we do nothing,then we cannot use podman driver on RHEL7 unfortunately.

BR,
Piotr Richter

mikenomitch:
Hey @rikislaw,

Sorry about the delay on this, we’ve been discussing this internally. 

As we mentioned before, we don’t feel 100% comfortable with the tradeoffs of this feature: the fact that a job can use more than its allotted resources, and that this could break things like quotas. So we wanted to nudge you one more time to look for a workaround on your end. But we realize that people aren't always in complete control of their infrastructure, so we want to provide a path forward.

If we do merge a PR like this, we’ll need to put it behind an opt-in config flag to discourage others from using the driver like this. We aren’t as worried about your specific case, since you’re part of this discussion and know the tradeoffs, but we want to make sure others don’t inadvertently break their quotas and resource restrictions.

So, assuming you want to go forward with this, I think there are two options:
1. You can use the PR as is in a forked version of the driver.
2. You can add the config flag conditions to this, and then we would merge the PR.

Option 1 is a bit less coding on your part, so that might be preferable.

If you go with Option 2, the rough path forward would be: 
* adding a bool to https://github.com/hashicorp/nomad-driver-podman/blob/main/config.go#LL126C4-L126C4
* adding driverConfig.BypassResourceControlls (or whatever we call it) to the conditionals from the PR
* adding it to the docs (https://github.com/hashicorp/nomad-driver-podman#driver-configuration) with a line that discourages its use.
* We would review and merge

We appreciate your patience on this and the work you’ve already done. Hopefully one of those options works for you!

shoenig:
It isn't possible to correctly support rootless operation with cgroups v1. Making it "work" by bypassing resource isolation isn't a solution, and not something we will support.

PR #206, opened 2023-01-10
Entrypoint and volumes improvement

rikislaw:
changes:
1.  SELinux label is not added by default to all volumes/mounts when we specify usage of SELinux label in main Nomad config for Podman driver. This setting goes only to standard Nomad mounts. For volumes defined in job definition, we need to provide `z` label for each volume, where we want to use it.  With previous approach I was not able to mount share, that didn't require providing `z` label at all.  I had this issue on RHEL7/8 with enabled selinux and for NFS mounts. NFS mounts were not working (permission denied) when mounted with `z` label.
2.  Entrypoint works exactly the same as for Docker driver. I changed it mainly because with current approach I was not able to disable entrypoint provided in the image and use only CMD.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ Ensure that Nomad can find the plugin, see [plugin_dir](https://www.nomadproject
 * volumes stanza:
 
   * enabled - Defaults to true. Allows tasks to bind host paths (volumes) inside their container.
-  * selinuxlabel - Allows the operator to set a SELinux label to the allocation and task local bind-mounts to containers. If used with _volumes.enabled_ set to false, the labels will still be applied to the standard binds in the container.
+  * selinuxlabel - Applies SELinux label to the standard binds in the container. To apply it also for additional defined volumes, `z` options needs to be applied in config->volumes section.
 
 ```hcl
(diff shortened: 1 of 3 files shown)


lgfa29 (changes requested):
Hi @rikislaw 👋 

Thank you for the PR! 

Matching the entrypoint behaviour to the Docker task driver makes sense to me. But this would be breaking change since it modifies the type of an existing driver configuration, so we'll need to be a bit more careful merging and releasing this change.

As for the SELinux label change, this was suggested before in https://github.com/hashicorp/nomad-driver-podman/pull/66 but there were some concerns about the impact on CSI volumes.

Would you mind splitting this PR into two so we can manage the changes separately? 

Thanks!

rikislaw:
Hi @Igfa29, thanks for comments, I created two new PRs:
https://github.com/hashicorp/nomad-driver-podman/pull/208
https://github.com/hashicorp/nomad-driver-podman/pull/209

Regarding Entrypoint PR, I will create also alternative solution in another PR. This one will be not breaking.

I'm closing this PR.

PR #209, opened 2023-01-16
Podman entrypoint works the same as for docker driver

rikislaw:
Entrypoint works exactly the same as for Docker driver. With previous approach I was not able to disable entrypoint provided in the image and use only custom CMD.

Alternative solution here: https://github.com/hashicorp/nomad-driver-podman/pull/210


lgfa29:
@towe75 I think this may be the first breaking change we had in a while, but it also seems like a good change to have. 

I was thinking of releasing a v0.4.2 with what we have merge already and then a v0.5.0 with this PR. 

What do you think? Do you have any breaking changes you were planning on releasing?

lgfa29 (approved):
Good news @rikislaw, @schmichael noted that changing the type would cause existing tasks to fail when using the new driver, and there's a chicken-and-egg problem that you can't update your job to the list syntax until you update the task driver.

Luckily we can use the `any` type and then cast the value depending on what the user provides, so we can now support both a list and a string, so this won't be a breaking change 🎉 

I pushed a couple of commits to implement this and also added a deprecation warning if a task starts with just a string. We still want to nudge people towards the new syntax 😅

schmichael (approved):

PR #210, opened 2023-01-16
entrypoint supports additional arguments

rikislaw:
Alternative solution to: https://github.com/hashicorp/nomad-driver-podman/pull/209

In this change, we pass entrypoint as list, when it is defined with whitespaces, eg. "/bin/bash -c" will be passed as ["/bin/bash","-c"] to podman api.  With this change, we can overwrite existing Entrypoint and Command with some new Command and neutralize image defined Entrypoint.

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -392,7 +392,7 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 	allArgs = append(allArgs, driverConfig.Args...)
 
 	if driverConfig.Entrypoint != "" {
-		createOpts.ContainerBasicConfig.Entrypoint = append(createOpts.ContainerBasicConfig.Entrypoint, driverConfig.Entrypoint)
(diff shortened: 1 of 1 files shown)


lgfa29:
Thanks for splitting the PR @rikislaw!

Unfortunately this is still a breaking change. For example, this job behaves differently depending on the plugin binary being used:

```hcl
job "example" {
  datacenters = ["dc1"]

  group "cache" {
    task "redis" {
      driver = "podman"

      config {
        image      = "docker://redis:7"
        entrypoint = "ls -a /"
      }
    }
  }
}
```

Granted, this is an unrealistic example 😅 but it's an unexpected behaviour change nevertheless that could surprise users.

Since they are both breaking changes, I think it would be better to go with https://github.com/hashicorp/nomad-driver-podman/pull/209 since it's the direction we want to go anyway.

So I'm going to close this one in favour of #209 and we can continue from there.

Thanks again for the PRs!

PR #212, opened 2023-02-03
NOMAD-236 userns podman configuration option

acornies:
I haven't been able to run the tests locally, but I want to get a general sense on if this is the right approach or not.

- add userns to task config
- add parsing for userns config option
- add the driver configuration, uncomment api portion
- tests


driver_test.go
```diff
@@ -1485,6 +1485,24 @@ func TestPodmanDriver_ReadOnlyFilesystem(t *testing.T) {
...
+}
+
+// check userns mode configuration (parsed value)
+func TestPodmanDriver_UsernsModeParsed(t *testing.T) {
+	taskCfg := newTaskConfig("", busyboxLongRunningCmd)
+	taskCfg.UserNS = "keep-id:uid=200,gid=210"
+	inspectData := startDestroyInspect(t, taskCfg, "userns")
+
+	require.Equal(t, "keep-id:uid=200,gid=210", inspectData.HostConfig.UsernsMode)
+}
```
lgfa29, line 1504: I'm not sure why the Github Action didn't run (it does this some times 😞), but I ran it locally and I think these need to run with Podman in rootless mode? The first test always returns an empty string:

```
--- FAIL: TestPodmanDriver_UsernsMode (6.28s)
    driver_test.go:1494:
                Error Trace:    /Users/laoqui/go/src/github.com/hashicorp/nomad-driver-podman/driver_test.go:1494
                Error:          Not equal:
                                expected: "host"
                                actual  : ""

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -host
                                +
                Test:           TestPodmanDriver_UsernsMode
FAIL
FAIL    github.com/hashicorp/nomad-driver-podman        6.313s
FAIL
```

The second test says the mode is not supported:

```
--- FAIL: TestPodmanDriver_UsernsModeParsed (4.11s)
    driver_test.go:2253:
                Error Trace:    /Users/laoqui/go/src/github.com/hashicorp/nomad-driver-podman/driver_test.go:2253
                                                        /Users/laoqui/go/src/github.com/hashicorp/nomad-driver-podman/driver_test.go:1501
                Error:          Received unexpected error:
                                rpc error: code = Unknown desc = failed to start task, could not create container: cannot create container, status code: 500: {"cause":"keep-id is only supported in rootless mode","message":"keep-id is only supported in rootless mode","response":500}
                Test:           TestPodmanDriver_UsernsModeParsed
```

I think our tests run as `root` so we may need to skip these in CI.

lgfa29 (changes requested):
Hi @acornies 👋 

Thanks for the PR! I left some comments for further discussion. You mentioned not being able to run the tests, is there any problem in particular?

driver.go
```diff
@@ -496,6 +496,21 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 	createOpts.ContainerSecurityConfig.ReadOnlyFilesystem = driverConfig.ReadOnlyRootfs
 	createOpts.ContainerSecurityConfig.ApparmorProfile = driverConfig.ApparmorProfile
 
+	// Populate --userns mode only if configured
+	if len(driverConfig.UserNS) > 0 {
+		userns := strings.Split(driverConfig.UserNS, ":")
```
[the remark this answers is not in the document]
acornies: Good point and nice catch. My understanding is that the api.Namespace value should be able to have what you stated above.

driver.go
```diff
@@ -667,6 +682,29 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
...
+		api.DefaultKernelNamespaces: {},
+	}
+
+	m := api.NamespaceMode(s)
+	_, ok := mode[m]
+	if !ok {
+		return c, fmt.Errorf(`cannot parse:[%s] as userns (namespace mode)`, s)
+	}
+	return m, nil
+}
```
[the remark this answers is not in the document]
acornies: Understood. I didn't know if we wanted to validate these things up front, but I'm all for deleting the hardcoded section for now.

driver_test.go
acornies: Part of the problem is I'm running a Macbook M1 so let me figure out...

lgfa29: Same 😅 

I've been using [Lima](https://github.com/lima-vm/lima) with this configuration, and so far it has been working well:
https://gist.github.com/lgfa29/f24eec7bb39f4bda28d7765bec8740cc#file-podman-yaml

I think you may need to run the tests as `sudo`.

acornies:
Thanks @lgfa29 you beat me to it (skipping the tests). I don't think there's anything else atm. It's an effort to iterate on additional config options that would be useful to podman users. The referenced internal issue has some other stuff that might be useful but those can happen in another PR.

lgfa29:
Cool, thanks! Feel free to open more PRs in the future as needed 😄

PR #215, opened 2023-02-25
Allow for an `extra_labels` stanza similar to the Docker driver

rbarry82:
Log collection in observability stacks (such as [via Vector](https://github.com/hashicorp/nomad-pack-community-registry/tree/main/packs/vector)) when using the `podman` driver isn't as literate as Docker. 

Add the option to add extra labels so promtail/vector/filebeat/whatever can get a view of what Nomad-related things running Podman containers may be doing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 IMPROVEMENTS:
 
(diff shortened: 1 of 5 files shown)


lgfa29 (approved):
This is great, thank you for the PR @rbarry82!

I pushed a commit to remove an extra `defer` call from test and also to add a `CHANGELOG.md` entry.

PR #218, opened 2023-03-11
driver: fix image_pull_timeout

### LINKED ISSUE #207: Default `image_pull_timeout` exceeds `client_http_timeout`
The default `image_pull_timeout` of 5 minutes exceeds the default `client_http_timeout` of 60 seconds. The **actual** image pull timeout is effectively 60s by default.

I found it a bit confusing that these defaults contradict the docs: _image_pull_timeout ... cannot be longer than the client_http_timeout_. While debugging an issue where my image pulls were timing out on my slow Internet connection, I realized the client timeout needed to be bumped to match the pull timeout.

I'm trying to understand the reasoning for these defaults? Is it to establish a reasonable image timeout in the event someone raises the client timeout? I would propose that the client timeout is raised to match the image timeout. I'm happy to open a PR if this is makes sense? Thank you!

lgfa29:
Previously the value for `image_pull_timeout` was being capped by `client_http_timeout` because the image pull request used the same HTTP client as other requests.

Increasing `client_http_timeout` in order to be able to set a higher `image_pull_timeout` is not a good option since it requires updating all client's configuration. It also affects all other HTTP requests, which may not be desirable.

The change creates a separate client for slow operations that don't have a timeout configured. Operation timeout is controlled using a ContextWithTimeout().

Closes #207

### DIFF
--- a/driver_test.go
+++ b/driver_test.go
@@ -7,6 +7,11 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"net/http/httputil"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -72,6 +77,15 @@ func podmanDriverHarness(t *testing.T, cfg map[string]interface{}) *dtestutil.Dr
 
 	baseConfig := base.Config{}
 	pluginConfig := PluginConfig{}
+
+	// Set ClientHttpTimeout before calling SetConfig() because it affects the
+	// HTTP client that is created.
+	if v, ok := cfg["ClientHttpTimeout"]; ok {
+		if sv, ok := v.(string); ok {
+			pluginConfig.ClientHttpTimeout = sv
+		}
+	}
+
 	if err := base.MsgPackEncode(&baseConfig.PluginConfig, &pluginConfig); err != nil {
 		t.Error("Unable to encode plugin config", err)
 	}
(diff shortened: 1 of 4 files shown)


driver_test.go
```diff
@@ -1914,6 +1928,91 @@ func startDestroyInspectImage(t *testing.T, image string, taskName string) {
...
+// driver's client_http_timeout.
+//
+// This test starts a proxy on port 5000 and requires the machine running the
+// test to allow an insecure registry at localhost:5000. To run this test add
+// the following to /etc/containers/registries.conf:
+//
+// [[registry]]
+// location = "localhost:5000"
+// insecure = true
+func TestPodmanDriver_Pull_Timeout(t *testing.T) {
```
lgfa29, line 1942: I have no idea how to test this reliably 🤷 

I couldn't find any documentation on how to make a Podman API request to ignore insecure TLS so this needs to be configured in the host.

courtland:
lgtm. Thank you.

driver_test.go
jdoss: This is the correct way to do an insecure registry. I use `insecure = true` when firing up an ephemernal registry for local development.

jdoss (approved):

tgross (approved):
LGTM

driver_test.go
lgfa29: Thanks! I was looking for a way to test this that wouldn't require a specific setting on the local machine, but I couldn't figure out a way, so that's the best we have for now 😅

PR #235, opened 2023-04-12
deps: update repository dependencies and cleanup linters

shoenig:
This PR bulk updates dependencies for the podman driver. It adds a `replace` statement for `nomad`, which includes updates to driver plumbing that we want, which will ship in 1.6. 

It also cleans up the golangci linter file, removing dead linters and adding some fresh ones. Also cleans up the source code in a few places to pass the linter checks. 

Also bumps Go version to 1.20.3


.golangci.yml
```diff
@@ -2,83 +2,74 @@
...
+    - scripts
 
-  # which files to skip: they will be analyzed, but issues from them
-  # won't be reported. Default value is empty list, but there is
-  # no need to include all autogenerated files, we confidently recognize
-  # autogenerated files. If it's not please let us know.
+  # Skip files
   skip-files:
     - ".*\\.generated\\.go$"
     - ".*bindata_assetfs\\.go$"
```
lgfa29, line 21: I think these are copy-pasta from Nomad and could be removed as well. No harm in keeping them thought 🤷

driver.go
```diff
@@ -493,16 +494,17 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 	createOpts.ContainerBasicConfig.Labels = driverConfig.Labels
 
 	// Logging
-	if driverConfig.Logging.Driver == "" || driverConfig.Logging.Driver == LOG_DRIVER_NOMAD {
+	switch {
+	case driverConfig.Logging.Driver == "" || driverConfig.Logging.Driver == LOG_DRIVER_NOMAD:
```
lgfa29, line 498:
```suggestion
	switch driverConfig.Logging.Driver {
	case "", LOG_DRIVER_NOMAD:
```

Maybe? Seems simpler like this.

driver.go
```diff
@@ -547,9 +549,9 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 	}
 
 	if driverConfig.MemorySwap != "" {
-		swap, err := memoryInBytes(driverConfig.MemorySwap)
-		if err != nil {
-			return nil, nil, err
+		swap, memErr := memoryInBytes(driverConfig.MemorySwap)
+		if memErr != nil {
+			return nil, nil, memErr
```
lgfa29, line 554: Is using `err` considered variable shadowing in cases like this?

lgfa29 (approved):
Thanks for the cleanup!

driver.go
shoenig: ooh nice catch! i was just bulldozing my way through these >_<

driver.go
shoenig: Yup; I'm not sure being this strict about shadowing will turn out for the best, but figured I'd give a try while doing some work on this repo

PR #239, opened 2023-04-13
tests: detect CI without using nomad import

shoenig:
This PR

- adds a `ci.Parallel()` helper
- calls that instead of `testutils.IsCI()` from the `nomad` package
- runs most tests in parallel by default now
- replaces some duration calculation with direct values instead of calling `testutils.TestMultiplier()`
- add an explicit check for whether SELinux tests should run (`$CI_SELINUX`)

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -81,7 +81,6 @@ test: ## Run unit tests
 	go test -v -race ./...
 
 .PHONY: test-ci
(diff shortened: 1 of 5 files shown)


lgfa29 (approved):
I thought some tests would fail if they ran in parallel, but apparently not, so that's good 😄

PR #247, opened 2023-05-16
build: update deprecated GitHub Actions

tgross:
Many of the GitHub Actions from the build pipeline are written in a truly ancient version of NodeJS. Upgrade to more recent versions.

Fixes: #246


lgfa29 (approved):
Just pushed a commit to handle a breaking change we noticed in the nightly release (ref: https://github.com/hashicorp/nomad-autoscaler/pull/635)

jrasell (approved):
LGTM!

tgross:
Pushed a commit to remove RelEng from the codeowners file as well.

PR #248, opened 2023-05-18
cgroups: set nomad.slice as the cgroup parent

shoenig:
Change ownership of the cgroup created by libpod to be Nomad's nomad.slice.

Still need to figure out what to do about the randomized cgroup name
being used by libpod, but that's another PR.


driver.go
```diff
@@ -765,6 +768,13 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 	return handle, driverNet, nil
 }
 
+// setupCgroup customizes where the cgroup lives (v2 only)
+func (d *Driver) setupCgroup(opts api.SpecGenerator) {
+	if d.cgroupV2 {
+		opts.ContainerCgroupConfig.CgroupParent = "nomad.slice"
```
lgfa29, line 774: I think it's pretty safe to assume that `nomad.slice` would never change right? So we don't need to make this configurable.

lgfa29 (approved):

driver.go
shoenig: Yeah it's fine for now. I'm working on an RFC that completely overhauls Nomad's cgroup management, but that won't come about until 1.7

PR #255, opened 2023-06-12
driver: add support for configuring extra_hosts in podman tasks

### LINKED ISSUE #244: [Podman] Support task.config.extra_hosts also for Podman Driver
### Proposal
We have currently a feature difference between the Docker driver and the Podman driver even tho the Docker driver does not fully support it in the right way it would be nice if we can use the `--add-host=host:ip` command is used when we use the config.ext

### Use-cases
As a maintainer, I want to route specific URLS (often .cloud URLs) to my VPN/NAT routing because of the security of my network.
fullygenerated.url.aws.service.cloud should go to IP 172.168.1.1

### Attempted Solutions
We seem to be able to do this in the Docker setup but not in the Podman setup:

[Docker documentation](https://developer.hashicorp.com/nomad/docs/drivers/docker#extra_hosts)
This is fully supported for the Podman https://docs.podman.io/en/latest/markdown/options/add-host.html and is simular variables see https://docs.podman.io/en/latest/markdown/options/add-host.html

shoenig:
This PR adds support for setting extra_hosts on in podman task configuration,
bringing the podman driver support up to par with the docker driver.

Closes #244

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -34,6 +34,7 @@ import (
 	pstructs "github.com/hashicorp/nomad/plugins/shared/structs"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/ryanuber/go-glob"
+	"golang.org/x/exp/slices"
 	"golang.org/x/sync/singleflight"
 )
 
@@ -652,6 +653,11 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 		}
 	}
 
+	// carefully add extra hosts (--add-host)
+	if extraHostsErr := setExtraHosts(driverConfig.ExtraHosts, &createOpts); extraHostsErr != nil {
+		return nil, nil, extraHostsErr
+	}
+
 	portMappings, err := d.portMappings(cfg, driverConfig)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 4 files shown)


driver.go
```diff
@@ -702,8 +708,8 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 
 	cleanup := func() {
 		d.logger.Debug("Cleaning up", "container", containerID)
-		if err = d.podman.ContainerDelete(d.ctx, containerID, true, true); err != nil {
-			d.logger.Error("failed to clean up from an error in Start", "error", err)
+		if cleanupErr := d.podman.ContainerDelete(d.ctx, containerID, true, true); cleanupErr != nil {
+			d.logger.Error("failed to clean up from an error in Start", "error", cleanupErr)
```
shoenig, line 712: fun fact - this would overwrite `err` during the `cleanup()` call, causing what is now `startErr` to always be nil, which was beyond confusing.

shoenig:
Spot check,

job file

```hcl
job "sleep" {
  group "group" {
    count = 1

    task "sleep" {
      driver = "podman"

      config {
        image = "docker.io/bash:5"
        args  = ["sleep", "infinity"]

        extra_hosts = ["test4.localhost:127.0.0.2", "test6.localhost:[::1]"]
      }

      resources {
        cores  = 2
        memory = 64
      }
    }
  }
}
```

exec into the container and cat `/etc/hosts`, see our extra hosts got added

```
➜ noamd alloc exec 28 /bin/sh
/ # cat /etc/hosts
127.0.0.1 localhost
127.0.1.1 diablo

# The following lines are desirable for IPv6 capable hosts
::1     ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
127.0.0.2 test4.localhost
[::1] test6.localhost
10.88.0.53      581de9968e4c sleep-28c28dd8-62e5-dfbc-6411-54101a77c90e
10.88.0.1 host.containers.internal
```

driver.go
lgfa29: Oof..nice catch!

lgfa29 (approved):
LGTM! Just needs a CHANGELOG entry 😄

PR #265, opened 2023-07-05
driver: support for credentials helper and auth config file

### LINKED ISSUE #132: Docker Registry Auth
Is it possible to configure a credentials file for podman to use. Similar to how docker auth can be configured:

```
    config {
        auth {
            config = "/path/to/docker/creds/file"
        }
    }
```
containing something like:

```
{
	"auths": {
		"docker.io": {
			"auth": "erfi7sYi89234xJUqaqxgmzcnQ2rRFWM5aJX0EC="
		},
		"quay.io": {
			"auth": "juQAqGmz5eR1ipzx8Evn6KGdw8fEa1w5MWczmgY="
		}
	}
}
```

podman itself supports `--authfile=path` and `REGISTRY_AUTH_FILE` env, it seems.

### LINKED ISSUE #174: Add option to point at auth.json file
I was trying to keep container registry credentials out of the Nomad job specification. My idea was to store them in Vault and use a Nomad template to pull them. As far as I can tell, Podman doesn't let you override using a set of environment variables - rather you point it at an `auth.json` file on the command line, or use the `REGISTRY_AUTH_FILE` environment variable to tell it where to read credentials from [see docs][pml].

I did `podman login` on a host and noted the generated `$XDG_RUNTIME_DIR/containers/auth.json` file.  I then grabbed the `auth` portion and stored it in Vault.

In the Nomad job specification I used a template to create a version of the `auth` file according to the [format][auth] as below:

```hcl
      template {
        data = <<EOH
        {
          "auths": {
            "ghcr.io": {
              "auth": "{{with secret "secret/data/github/creds"}}{{.Data.data.auth}}{{end}}"
            }
          }
        }
        EOH

        destination = "local/auth.json"
        change_mode = "noop" # We don't care if this value changes once we are running
      }
```

I then tried to get Podman to recognize that file using:

```hcl
      env {
        HTTP_PORT          = "${NOMAD_PORT_http}"
        ENV                = "dev"
        # So Podman can find the auth file
        REGISTRY_AUTH_FILE = "${NOMAD_TASK_DIR}/auth.json"
      }
```

Unfortunately, Podman would not pick up the credentials. I viewed the created file under `local` and it looked correct to me.

The tasks all failed because the container could not be pulled from the registry.

My guess is that `REGISTRY_AUTH_FILE` is not available to the driver, so it isn't using it?

The feature that would be useful is adding an option to point the driver at the credentials file to use.

[pml]: https://docs.podman.io/en/latest/markdown/podman-login.1.html

[auth]:https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md

shoenig:
This PR adds support for specifying an external credentials helper and/
  or an external credentials file (e.g. "auth.json"). The plugin configuration
  now has an `auth` block with fields `helper` and `config` (similar to
  the docker driver). We also now have an `auth_soft_fail` option in Task
  config for cases where someone has configured the `auth` block in plugin
  config, but has a task that is using a public image with no credentials.
  In that case setting `auth_soft_fail` is used to ignore the fact that
  no credentials will be found for the given public image. (This is also
  how the docker driver works, I didn't come up with this).

  Unlike the docker driver, the podman driver still does not support
  specifying a credentials helper _in_ the external "auth.json"
  credentials file. If there is demand for that use case we can add
  it, but in the short term it seems like just the plugin's support for
  specifying a credentials helper could be sufficient. And it's a lot of
  code to do the other thing.

Closes #132
Closes #174

Note: docs and changelog will be in a follow up PR (along with all the other 0.5.0 work).

e2e tests will be in https://github.com/hashicorp/nomad/commits/e2e-podman-private-registry
(still needs to tweaking to work with the actual e2e runner)

### DIFF
--- a/api/image_pull.go
+++ b/api/image_pull.go
@@ -10,52 +10,55 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+
+	"github.com/hashicorp/nomad-driver-podman/registry"
 )
 
 // ImagePull pulls a image from a remote location to local storage
-func (c *API) ImagePull(ctx context.Context, nameWithTag string, auth ImageAuthConfig) (string, error) {
-	var id string
-	headers := map[string]string{}
+func (c *API) ImagePull(ctx context.Context, pullConfig *registry.PullConfig) (string, error) {
+	pullConfig.Log(c.logger)
 
-	// handle authentication
-	usesAuth := auth.Username != "" && auth.Password != ""
-	if usesAuth {
-		authHeader, err := NewAuthHeader(auth)
-		if err != nil {
-			return "", err
-		}
-		headers["X-Registry-Auth"] = authHeader
+	var (
+		headers    = make(map[string]string)
+		repository = pullConfig.Repository
+		tlsVerify  = pullConfig.TLSVerify
+	)
+
+	auth, err := registry.ResolveRegistryAuthentication(repository, pullConfig)
+	if err != nil {
+		return "", fmt.Errorf("failed to determine authentication for %q", repository)
 	}
+	auth.SetHeader(headers)
+	c.logger.Info("HEADERS", "headers", headers)
 
-	c.logger.Trace("image pull details", "tls_verify", auth.TLSVerify, "reference", nameWithTag, "uses_auth", usesAuth)
+	urlPath := fmt.Sprintf("/v1.0.0/libpod/images/pull?reference=%s&tlsVerify=%t", repository, tlsVerify)
+	c.logger.Info("URL PATH", "urlPath", urlPath)
 
-	urlPath := fmt.Sprintf("/v1.0.0/libpod/images/pull?reference=%s&tlsVerify=%t", nameWithTag, auth.TLSVerify)
 	res, err := c.PostWithHeaders(ctx, urlPath, nil, headers)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("failed to pull image: %w", err)
 	}
-
 	defer ignoreClose(res.Body)
+
 	if res.StatusCode != http.StatusOK {
 		body, _ := io.ReadAll(res.Body)
 		return "", fmt.Errorf("cannot pull image, status code: %d: %s", res.StatusCode, body)
 	}
 
-	dec := json.NewDecoder(res.Body)
-	var report ImagePullReport
+	var (
+		dec    = json.NewDecoder(res.Body)
+		report ImagePullReport
+		id     string
+	)
 	for {
-		decErr := dec.Decode(&report)
-		if errors.Is(decErr, io.EOF) {
+		decodeErr := dec.Decode(&report)
+		if errors.Is(decodeErr, io.EOF) {
 			break
-		} else if decErr != nil {
-			return "", fmt.Errorf("Error reading response: %w", decErr)
-		}
-
-		if report.Error != "" {
-			return "", errors.New(report.Error)
-		}
-
-		if report.ID != "" {
+		} else if decodeErr != nil {
+			return "", fmt.Errorf("failed to read image pull response: %w", decodeErr)
+		} else if report.Error != "" {
+			return "", fmt.Errorf("image pull report indicates error: %s", report.Error)
+		} else if report.ID != "" {
 			id = report.ID
 		}
 	}
--- a/api/image_pull.go
+++ b/api/image_pull.go
@@ -20,19 +20,22 @@ func (c *API) ImagePull(ctx context.Context, pullConfig *registry.PullConfig) (s
 
 	var (
 		headers    = make(map[string]string)
-		repository = pullConfig.Repository
+		repository = pullConfig.Image
 		tlsVerify  = pullConfig.TLSVerify
 	)
 
-	auth, err := registry.ResolveRegistryAuthentication(repository, pullConfig)
-	if err != nil {
-		return "", fmt.Errorf("failed to determine authentication for %q", repository)
+	// if the task or driver are configured with an auth block, attempt to find
+	// credentials that are compatible with the given image, and set the appropriate
+	// header if found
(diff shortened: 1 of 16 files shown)


shoenig:
Spot check, the e2e suite in the branch above where we try each configuration against a private authenticated registry.

```
diablo nomad/e2e/podman on e2e-podman-private-registry
➜ go test -v
=== RUN   TestPodman
=== RUN   TestPodman/testRedis
=== RUN   TestPodman/testAuthBasic
=== RUN   TestPodman/testAuthFileStatic
=== RUN   TestPodman/testAuthHelper
--- PASS: TestPodman (88.28s)
    --- PASS: TestPodman/testRedis (13.04s)
    --- PASS: TestPodman/testAuthBasic (23.08s)
    --- PASS: TestPodman/testAuthFileStatic (28.10s)
    --- PASS: TestPodman/testAuthHelper (24.07s)
PASS
ok      github.com/hashicorp/nomad/e2e/podman   88.292s
```

registry/authentication.go
```diff
@@ -0,0 +1,185 @@
...
+		tls = pc.TLSVerify
+		username = r.Username
+		email = r.Email
+		server = r.ServerAddress
+		password = r.Password != ""
+		token = r.IdentityToken != ""
+	}
+
+	// todo: trace
+	logger.Info("pull config",
```
lgfa29, line 81:
```suggestion
	logger.Trace("pull config",
```
TODONE 😄

registry/fileconfig.go
```diff
@@ -0,0 +1,87 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+package registry
+
+import (
+	"encoding/json"
+	"os"
+)
+
+// [Glossery]
```
lgfa29, line 11:
```suggestion
// [Glossary]
```

registry/fileconfig.go
```diff
@@ -0,0 +1,87 @@
...
+// that stores or references credentials that enable authentication into specific
+// registries or even repositories.
+//
+// reference: https://www.mankier.com/5/containers-auth.json
+// alternate: https://man.archlinux.org/man/containers-auth.json.5
+type CredentialsFile struct {
+	Auths map[string]EncAuth `json:"auths"`
+
+	// CredHelpers is currently not supported by the podman task driver
+	// CredHelpers map[string]string  `json:"credHelpers"`
```
lgfa29, line 29: Just to make sure I understand things, these credential helpers are the same concepts as the [plugin config cred helper](https://github.com/hashicorp/nomad-driver-podman/pull/265/files?diff=split&w=0#diff-0e426a43248661127a0c0ee115aef7a1093b635f8993b3f7ebb1dd9f05b8f249R138) (a script that is executed to return a username and password), but are different because these are defined in the `auth.json` file, so in order to support them we will have to parse them from the JSON file and run something like [`authFromCredsHelper`](https://github.com/hashicorp/nomad-driver-podman/pull/265/files?diff=split&w=0#diff-bc2e0486c538e9f1c2b65ab5b564fbd9c2def7fd43bf85439d38a33ecf8d7ba4R44) on them?

lgfa29 (approved):
LGTM!

I don't know if it would be helpful in your case, but I have a test pattern when I create a proxy between the driver and a registry to manipulate requests 
https://github.com/hashicorp/nomad-driver-podman/blob/40566ed5b6c612a2542fb8d30d2844d3a04c22a8/driver_test.go#L1991-L2076

I think you could do something like that to check the auth header is being set properly but it's quite a bit of heavy lifting 😅

api/image_pull.go
```diff
@@ -10,54 +10,61 @@ import (
...
-func (c *API) ImagePull(ctx context.Context, nameWithTag string, auth ImageAuthConfig) (string, error) {
-	var id string
-	headers := map[string]string{}
-
-	// handle authentication
-	usesAuth := auth.Username != "" && auth.Password != ""
-	if usesAuth {
-		authHeader, err := NewAuthHeader(auth)
+func (c *API) ImagePull(ctx context.Context, pullConfig *registry.PullConfig) (string, error) {
+	pullConfig.Log(c.logger)
```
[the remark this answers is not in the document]
shoenig: Yeah it's a bit unusual but it nicely hides the ~20 lines of gunk specific to the thing being logged

api/image_pull.go
```diff
@@ -10,54 +10,61 @@ import (
...
+
+	var (
+		headers    = make(map[string]string)
+		repository = pullConfig.Image
+		tlsVerify  = pullConfig.TLSVerify
+	)
```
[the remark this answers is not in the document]
shoenig: Nope none at all; doing it this way just makes `gofmt` line them up nicely

registry/authentication.go
shoenig: :sweat_smile:

registry/fileconfig.go
shoenig: That's exactly correct!

The main advantage to implementing support for specifying helpers in the `auth.json` file is that you can directly correlate a registry/image to a specific helper.

However, you could also just implement your own credsHelper wrapper that does the same mapping, and invokes the correct helper given a specific registry.

jamjon3:
Just a thought as a _way_ to handle auth.json

In the client /etc/nomad.d/nomad.env

add

REGISTRY_AUTH_FILE=/etc/nomad.d/registry/auth.json

That's podman's way of handling an auth file as an environment variable. Just set it in the nomad.env that's already provided and you don't need to set it anywhere else.

bru-msholomov:
Where can I see an example of how to use it? I have not found any documentation on this

shoenig:
Oh that's a good point @bru-msholomov we should copy over some of the docs from the docker driver (this mechanism was copied from there and works the same). The docker docs are here: https://developer.hashicorp.com/nomad/docs/drivers/docker#authentication and should get you started.

Issue #227, opened 2023-04-03
nomad-driver-podman_0.4.2_linux_amd64.zip artifact has ELF architecture: EM_X86_64


jdoss:
I am finally upgrading to 0.4.2 and it seems the amd64 artifact is EM_X86_64. This impacts the nightly artifact too. Building from source results in the correct ELF architecture.

```
Apr 03 01:26:36 nomad[15887]:     2023-04-03T01:26:36.639Z [ERROR] agent: error starting agent:
Apr 03 01:26:36 nomad[15887]:   error=
Apr 03 01:26:36 nomad[15887]:   | failed to create plugin loader: failed to initialize plugin loader: failed to fingerprint plugins: 1 error occurred:
Apr 03 01:26:36 nomad[15887]:   | \t* Unrecognized remote plugin message:
Apr 03 01:26:36 nomad[15887]:   | This usually means
Apr 03 01:26:36 nomad[15887]:   |   the plugin was not compiled for this architecture,
Apr 03 01:26:36 nomad[15887]:   |   the plugin is missing dynamic-link libraries necessary to run,
Apr 03 01:26:36 nomad[15887]:   |   the plugin is not executable by this process due to file permissions, or
Apr 03 01:26:36 nomad[15887]:   |   the plugin failed to negotiate the initial go-plugin protocol handshake
Apr 03 01:26:36 nomad[15887]:   |
Apr 03 01:26:36 nomad[15887]:   | Additional notes about plugin:
Apr 03 01:26:36 nomad[15887]:   |   Path: /etc/nomad/plugins/nomad-driver-podman
Apr 03 01:26:36 nomad[15887]:   |   Mode: -rwxr-xr-x
Apr 03 01:26:36 nomad[15887]:   |   Owner: 0 [root] (current: 0 [root])
Apr 03 01:26:36 nomad[15887]:   |   Group: 0 [root] (current: 0 [root])
Apr 03 01:26:36 nomad[15887]:   |   ELF architecture: EM_X86_64 (current architecture: amd64)
Apr 03 01:26:36 nomad[15887]:   |
Apr 03 01:26:36 nomad[15887]:   |
Apr 03 01:26:36 nomad[15887]:   
```

shoenig:
Hi @jdoss, where did you get the zip archive from? I just checked the one from [releases](https://releases.hashicorp.com/nomad-driver-podman/0.4.2/) and it looks fine:

```
[...]

PR #266, opened 2023-07-06
deps: update nomad dependency to v1.6.0-beta.1

shoenig:
Should hopefully fix the CGO problem in https://github.com/hashicorp/nomad-driver-podman/issues/227


lgfa29 (approved):
I'm not sure how to test this, but if you need something to smoke test I built some binaries from this branch:
https://github.com/hashicorp/nomad-driver-podman/actions/runs/5480372129

shoenig:
@jdoss if you have access to the artifacts in ^, mind trying against Fedora CoreOS?

jdoss:
@shoenig I was able to test this out this evening on a client node and it worked as expected! Thank you for fixing this. 😄

ahjohannessen:
@shoenig Is a 0.4.3 with this fix soon released?

shoenig:
We'll do a Podman v0.5.0 release as soon as we have Nomad v1.6 GA, which should be in the next day or so.

PR #270, opened 2023-07-12
auth: skip registry auth file if config field is not set

### LINKED ISSUE #269: bug: need to ignore empty config field if auth block exists
In setting up the e2e test case, I found that if e.g. `auth.helper` is set but `auth.config` is not set, the task will fail to start with an empty file not found error. We need to skip checking the file config if the file field is empty.

shoenig:
Fixes #269 

No CL, the auth stuff hasn't shipped yet.


lgfa29 (approved):
Oh, nice catch!

Issue #277, opened 2023-08-04
Jobs are not correctly restarted on Podman 4.6.0


jonasdemoor:
# Description

We're running Nomad 1.5.3 with Podman 4.6.0 compiled from sources on Debian 11 (bullseye). I attached more details about our environment below.

This is how the state of Nomad and Podman looks like after starting the job:

```
$ nomad job status
ID            Type    Priority  Status   Submit Date
csi-nfs-node  system  50        running  2023-08-04T12:28:24+02:00

$ podman ps
CONTAINER ID  IMAGE                                         COMMAND               CREATED        STATUS        PORTS       NAMES
452b8b3dc9d0  registry.k8s.io/sig-storage/nfsplugin:v4.1.0  -v=5 --nodeid=nom...  6 seconds ago  Up 6 seconds              plugin-928c60c6-2c62-5e13-f10d-054ac74f16f4

$ journalctl -u podman
Aug 04 12:28:54 <redacted> podman[70819]: @ - - [04/Aug/2023:12:28:54 +0200] "GET /v1.0.0/libpod/containers/452b8b3dc9d04cfb4a41ee11f3590f46c75ce00d60fef9826340e6e577637ca3/stats?stream=false HTTP/1.1" 200 1112 "" "Go-http-client/1.1"
Aug 04 12:28:55 <redacted> podman[70819]: @ - - [04/Aug/2023:12:28:55 +0200] "GET /v1.0.0/libpod/containers/452b8b3dc9d04cfb4a41ee11f3590f46c75ce00d60fef9826340e6e577637ca3/stats?stream=false HTTP/1.1" 200 1112 "" "Go-http-client/1.1"
```

After stopping the container, Nomad lets Podman correctly stop the container, but doesn't gc/remove it. Even though the job is dead according to Nomad, it keeps trying to get stats for that container.

```
$ nomad job stop csi-nfs-node
==> 2023-08-04T12:30:15+02:00: Monitoring evaluation "21d85be1"
    2023-08-04T12:30:15+02:00: Evaluation triggered by job "csi-nfs-node"
    2023-08-04T12:30:16+02:00: Evaluation status changed: "pending" -> "complete"
==> 2023-08-04T12:30:16+02:00: Evaluation "21d85be1" finished with status "complete"
$ nomad job status           
ID            Type    Priority  Status          Submit Date
csi-nfs-node  system  50        dead (stopped)  2023-08-04T12:28:24+02:00

$ podman ps -a
[...]

PR #278, opened 2023-08-07
Adjust `ContainerStats` for Podmod 4.6.0 responses

sushain97:
Fixes https://github.com/hashicorp/nomad-driver-podman/issues/277

### DIFF
--- a/api/container_stats.go
+++ b/api/container_stats.go
@@ -49,7 +49,7 @@ func (c *API) ContainerStats(ctx context.Context, name string) (Stats, error) {
 
 	// Since podman 4.6.0, a 200 response with `container is stopped` is returned for stopped containers.
 	var errResponse Error
-	if err := json.Unmarshal(body, &errResponse); err == nil && errResponse.Cause == "container is stopped" {
+	if unmarshalErr := json.Unmarshal(body, &errResponse); unmarshalErr == nil && errResponse.Cause == "container is stopped" {
 		return stats, ContainerNotFound
 	}
 
--- a/api/container_stats.go
+++ b/api/container_stats.go
@@ -49,7 +49,7 @@ func (c *API) ContainerStats(ctx context.Context, name string) (Stats, error) {
 
 	// Since podman 4.6.0, a 200 response with `container is stopped` is returned for stopped containers.
 	var errResponse Error
-	if unmarshalErr := json.Unmarshal(body, &errResponse); unmarshalErr == nil && errResponse.Cause == "container is stopped" {
+	if _ := json.Unmarshal(body, &errResponse); errResponse.Cause == "container is stopped" {
 		return stats, ContainerNotFound
 	}
 
--- a/api/container_stats.go
+++ b/api/container_stats.go
@@ -49,7 +49,7 @@ func (c *API) ContainerStats(ctx context.Context, name string) (Stats, error) {
 
 	// Since podman 4.6.0, a 200 response with `container is stopped` is returned for stopped containers.
 	var errResponse Error
-	if _ := json.Unmarshal(body, &errResponse); errResponse.Cause == "container is stopped" {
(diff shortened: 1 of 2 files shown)


Procsiab:
I can report that this PR fixes #277 in my environment, on both x86_64 and aarch64 Nomad `1.6.1` clients with Podman `4.6.0-1`. Thanks for your quick response @sushain97 :muscle:

shoenig:
@sushain97 check the linter failure

```
==> Linting source code ...
Error: api/container_stats.go:52:5: shadow: declaration of "err" shadows declaration at line 22 (govet)
	if err := json.Unmarshal(body, &errResponse); err == nil && errResponse.Cause == "container is stopped" {
	   ^
make: *** [GNUmakefile:32: check] Error 1
```

sushain97:
@shoenig fixed

api/container_stats.go
```diff
@@ -47,6 +47,12 @@ func (c *API) ContainerStats(ctx context.Context, name string) (Stats, error) {
 		return stats, ContainerNotFound
 	}
 
+	// Since podman 4.6.0, a 200 response with `container is stopped` is returned for stopped containers.
+	var errResponse Error
+	if unmarshalErr := json.Unmarshal(body, &errResponse); unmarshalErr == nil && errResponse.Cause == "container is stopped" {
+		return stats, ContainerNotFound
+	}
```
lgfa29, line 54:
```suggestion
	if _ := json.Unmarshal(body, &errResponse); errResponse.Cause == "container is stopped" {
		return stats, ContainerNotFound
	}
```

I think `unmarshalErr` is irrelevant here? Since `errResponse` is not a pointer and `Cause` is just a string, in case of unmarshelling errors `Cause` would be an empty string.

https://go.dev/play/p/qhoq5l0pYIA

sushain97:
@lgfa29 does this look good?

api/container_stats.go
lgfa29: > However, I don't generally skip error checking.

That's indeed a very good practice 😄 

I hope this doesn't come back to hunt me, but in this case I think it's OK because we're doing anything with it and...

>  In this case, `errResponse` could be filled with partial data

True, but then we would either have the error message in `Cause` or an empty string right? Or maybe I missing another possible scenario? 😅 

> and in general its value is undefined when err != nil.

Not in this case since `errResponse` is a struct, not a pointer, so we would have all fields be zero values.

lgfa29 (approved):
This looks good to me, thanks for the PR @sushain97!

I'm deferring the final merge to @shoenig since he's been working more closely with the Podman driver more recently

api/container_stats.go
sushain97: > True, but then we would either have the error message in Cause or an empty string right? Or maybe I missing another possible scenario? 😅

In practice, I doubt that you are missing anything.

But I think about this with my "unspecified behavior" hat on. If the `Unmarshal` function doesn't say that it will leave the second argument alone if it returns an a non-nil error, I make no assumptions about what it might put there. It could just as well decide that filling in the `"container is stopped"` string in a `Cause` field is something it'll do. And, I couldn't really complain since it returned a non-nil error and didn't say that it will leave the second argument in any particular modified/unmodified state if it does that. I suspect that a failure mid-parse could actually modify the second argument (although, not in a way that I'd expect would change the behavior of this code).

But, I'm being overly pedantic :)

shoenig:
@sushain97 I think your fix has a compilation bug: 

```
if _ := json.Unmarshal(body, &errResponse); ...{
```

needs to be

```
if _ = json.Unmarshal(body, &errResponse); ... {
```

sushain97:
@shoenig fixed - the suggestion wasn't quite right.

Would be helpful to have workflows approved for the entire branch/PR :)

shoenig (approved):
LGTM; thanks @sushain97 !

sushain97:
No problem! Would it be possible to cut a 0.5.1 release with this fix?

PR #280, opened 2023-08-27
Pass nomad.slice as CGroup parent only on SystemD

### LINKED ISSUE #274: systemd slice received as cgroup parent when using cgroupfs: invalid argument
I get the following error with version 5.0.0 of this driver, but not 0.4.2:

```
rpc error: code = Unknown desc = failed to start task, could not create container: cannot create container, status code: 500: {"cause":"invalid argument","message":"systemd slice received as cgroup parent when using cgroupfs: invalid argument","response":500} 
```

Here is a sample job that doesn't work for me on the new version:

```hcl
job "example-1" {
  datacenters = ["dc1"]

  group "hello" {
    task "hello" {
      driver = "podman"

      config {
        image   = "redis:3.2"
        command = "/bin/bash"
        args    = ["-c", "echo 'Hello World!' && sleep infinity"]
      }

      resources {
        cpu    = 2
        memory = 20
      }
    }
  }
}
```

My client nodes are running podman version 4.5.1 on alpine 3.18. I am using cgroups v2. I tried running a patched build but my in-driver logging didn't give enough detail and I got lost after we call `ContainerCreate()`. Speaking of, I think this Request struct may be unused? https://github.com/hashicorp/nomad-driver-podman/blob/2f2353c4e76f5342a97e113645730dfedbdf54fe/api/container_create.go#L49-L75

I was a little suspicious of this commit: https://github.com/hashicorp/nomad-driver-podman/commit/c7c9e97c514b4ec80eadc50f731472326d3941c5, but I didn't bisect or try everything before that one just yet. I am happy to test out any patches to fix this issue! Fortunately I only wanted to upgrade for the new `tls_verify` support, but I can do without that for now.

Procsiab:
Let Nomad Driver Podman work again in an environment without SystemD, for example on Alpine Linux with `cgroupfs` as alternative CGroup manager.

Fixes #274

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -124,6 +124,8 @@ type Driver struct {
 	systemInfo api.Info
 	// Queried from systemInfo: is podman running on a cgroupv2 system?
 	cgroupV2 bool
+	// Queried from systemInfo: name of the cgroup manager
+	cgroupMgr string
 
 	// singleflight group to prevent parallel image downloads
 	pullGroup singleflight.Group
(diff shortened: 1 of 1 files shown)


shoenig (approved):
LGTM; thanks @Procsiab!

PR #285, opened 2023-09-29
Added a plugin logging configuration to specify defaults.

apollo13:
This is similar to https://github.com/hashicorp/nomad/pull/10156/files and allows for a configuration like:

```
    config {
	extra_labels = ["job_name", "job_id", "task_group_name", "task_name", "namespace", "node_name", "node_id"]
	logging {
		driver = "journald"
		options = {
			tag = "{{index .Config.Labels \"com.hashicorp.nomad.alloc_id\" }}"
		}
	}
    }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,11 @@ plugin "nomad-driver-podman" {
 }
 ```
 
+* logging stanza:
+
+  * type - Defaults to `"nomad"`. See the task configuration for details.
+  * options - Defaults to `{}`. See the task configuration for details.
+
 * client_http_timeout (string) Defaults to `60s` default timeout used by http.Client requests
 
 ```hcl
@@ -342,11 +347,9 @@ Ensure you're running Podman 3.1.0 or higher because of bugs in older versions.
 config {
   logging = {
     driver = "journald"
-    options = [
-      {
-        "tag" = "redis"
-      }
-    ]
+    options = {
+      "tag" = "redis"
+    }
   }
 }
 ```
--- a/config.go
+++ b/config.go
@@ -146,6 +146,11 @@ type TaskLoggingConfig struct {
 	Options hclutils.MapStrStr `codec:"options"`
 }
 
+// Empty returns true if the logging configuration is not set.
+func (l *TaskLoggingConfig) Empty() bool {
+	return l.Driver == "" && len(l.Options) == 0
+}
+
 // VolumeConfig is the drivers volume specific configuration
 type VolumeConfig struct {
 	Enabled      bool   `codec:"enabled"`
--- a/config.go
+++ b/config.go
@@ -44,6 +44,16 @@ var (
 		),
 		// optional extra_labels to append to all tasks for observability. Globs supported
 		"extra_labels": hclspec.NewAttr("extra_labels", "list(string)", false),
+
+		// logging options
+		"logging": hclspec.NewDefault(hclspec.NewBlock("logging", false, hclspec.NewObject(map[string]*hclspec.Spec{
+			"driver":  hclspec.NewAttr("driver", "string", false),
+			"options": hclspec.NewBlockAttrs("options", "string", false),
+		})), hclspec.NewLiteral(`{
+			driver = "nomad"
+			options = {}
+		}`)),
+
 		// the path to the podman api socket
 		"socket_path": hclspec.NewAttr("socket_path", "string", false),
 		// disable_log_collection indicates whether nomad should collect logs of podman
@@ -124,8 +134,8 @@ type GCConfig struct {
 
 // LoggingConfig is the tasks logging configuration
 type LoggingConfig struct {
-	Driver  string             `codec:"driver"`
-	Options hclutils.MapStrStr `codec:"options"`
+	Driver  string            `codec:"driver"`
+	Options map[string]string `codec:"options"`
 }
 
 // VolumeConfig is the drivers volume specific configuration
@@ -150,6 +160,7 @@ type PluginConfig struct {
 	ClientHttpTimeout    string           `codec:"client_http_timeout"`
 	ExtraLabels          []string         `codec:"extra_labels"`
 	DNSServers           []string         `codec:"dns_servers"`
+	Logging              LoggingConfig    `codec:"logging"`
 }
 
 // LogWarnings will emit logs about known problematic configurations
--- a/config.go
+++ b/config.go
@@ -132,10 +132,18 @@ type GCConfig struct {
 	Container bool `codec:"container"`
 }
 
-// LoggingConfig is the tasks logging configuration
+// LoggingConfig is the driver logging configuration
+// keep in sync with `TaskLoggingConfig`
 type LoggingConfig struct {
 	Driver  string            `codec:"driver"`
 	Options map[string]string `codec:"options"`
+}
+
+// LoggingConfig is the tasks logging configuration
+// keep in sync with `LoggingConfig`
+type TaskLoggingConfig struct {
+	Driver  string             `codec:"driver"`
+	Options hclutils.MapStrStr `codec:"options"`
 }
 
 // VolumeConfig is the drivers volume specific configuration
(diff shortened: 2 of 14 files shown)


config.go
```diff
@@ -123,8 +133,8 @@ type GCConfig struct {
 
 // LoggingConfig is the tasks logging configuration
 type LoggingConfig struct {
-	Driver  string             `codec:"driver"`
-	Options hclutils.MapStrStr `codec:"options"`
+	Driver  string            `codec:"driver"`
+	Options map[string]string `codec:"options"`
```
apollo13, line 137: I needed to switch this to a map from `MaptStrStr` because otherwise message pack would fail in `setConfig`:
```
    2023-09-29T12:25:52.047+0200 [ERROR] agent: error starting agent:
  error=
  | failed to create plugin loader: failed to initialize plugin loader: parsing plugin configurations failed: 1 error occurred:
  | \t* plugin "podman" (driver): failed to dispense plugin: setting config for plugin "podman" (driver) failed: rpc error: code = Unknown desc = SetConfig failed: msgpack decode error [pos 117]: invalid byte descriptor for decoding bytes, got: 0x74
  | 
```

not exactly sure why this happens :/

driver_test.go
```diff
@@ -516,6 +516,55 @@ func TestPodmanDriver_logNomad(t *testing.T) {
 	must.StrContains(t, stdoutLog, stderrMagic)
 }
 
+// Check default plugin logger
+// TODO: Can we make this nicer?
```
apollo13, line 520: This just tests that the default is set, preferably I'd like to provide a special `PluginConfig` that explicitly sets this value?

README.md
```diff
@@ -342,11 +347,9 @@ Ensure you're running Podman 3.1.0 or higher because of bugs in older versions.
 config {
   logging = {
     driver = "journald"
-    options = [
```
apollo13, line 345: Was this really supposed to be a list?

config.go
lgfa29: Hum...I'm not sure why this would happen either. But I'm a bit worried about backwards compatibility of changing this value type as msgpack can be a bit fuzzy with types.

I'm als worried about the godoc for `MapStrInt`. It seems like this struct was created to support an unconventional syntax that was used in older version of Nomad: https://github.com/hashicorp/nomad/commit/afb0c5ac452ba3bcf5c12907351dccb4edbc368b.

It's also interesting that the PR mentions that external plugins should not use this, but we don't document this anywhere. First time I hear about it 🙃 

> This is only required for built-in types that have backward
> compatibility constraints.  External drivers should use `BlockAttrs`
> instead, as they see fit.

But, since this is already like this, modifying it would be a breaking change.

Instead, I think it may be better to create a new struct just like this but for driver config instead of task. It's kind of annoying that both need to be kept in-sync (and we should note this in their godoc string), but that's already kind the case since `taskConfigSpec` and `configSpec` would need to change as well.

driver_test.go
lgfa29: I think this would be fine for now. In the future it would be interesting to turn https://github.com/hashicorp/nomad-driver-podman/blob/main/api/api.go into an interface and inject a mock into the driver that doesn't actually calls Podman and allows us to inspect the config the driver submitted to `ContainerCreate`.

But that's a whole different work stream 😅 

If you do feel like adding more through testing, another scenario to test would be to ensure the task config overwrites the plugin config.

This could be done with a test similar to this, but the plugin is configured with `journald` and the task with `nomad`. `nomad` should be picked as the correct log driver.

README.md
lgfa29: Hum...good point, it seems like it's supposed to be map 👍

lgfa29 (changes requested):
Thanks for the PR @apollo13!

And apologies for the delay in reviewing it. I'm slowly catching up to all the Ecosystem projects 😅 

The main change here would be to avoid modifying the task config type as it causes a backwards incompatibility.

Could you also add an entry to `CHANGELOG.md`?

Let me know if you have any questions!

apollo13:
Hi @lgfa29, thank you for your review. Can you please explain to me why changing the type affects backwards compat (trying to learn more about how nomad works)?

lgfa29:
> Can you please explain to me why changing the type affects backwards compat (trying to learn more about how nomad works)?

Sure! I should preface this by saying that I'm mostly going from the PR description on https://github.com/hashicorp/nomad/pull/5321 which is where `MapStrStr` was introduced.

HCL supports two syntax: native HCL and JSON. HCLv1 is somewhat under-specified in some senses, which create a few ambiguities, meaning two different HCL/JSON files may be decoded into the same struct.

https://github.com/hashicorp/nomad/pull/5321 lists some examples:

```
// hcl block
port_map {
  http = 80
  https = 443
}

// hcl assignment
port_map = {
  http  = 80
  https = 443
}

// json single element array of map (default in API response)
{"port_map": [{"http": 80, "https": 443}]}

// json array of individual maps (supported accidentally iiuc)
{"port_map: [{"http": 80}, {"https": 443}]}
```

HCLv2 is more strict, and makes a block (`port { ... }`) and a map assignment (`port = { ... }`) to be two different things. So, in order to support job files written with the old syntax, `MapStrStr` was created to manually decoded them the same way.

In this scenario, this means that a Podman driver written like this (note the missing `=` after `options`) would no longer work:

```hcl
task "..." {
  driver = "podman"

  config {
    logging = {
      driver = "journald"
      options {
        "tag" = "redis"
      }
    }
  }
  # ...
}
```

But it's probably worth testing it by running a task like the above using the latest driver version available and then comparing it to this PR.

It would also be interesting to test an upgrade path.

1. Start non-dev Nomad cluster (need to retain state).
2. Run a Podman job with a task like the above.
3. Stop Nomad client.
4. Update Podman driver with this PR.
5. Start Nomad client and verify if the alloc was able to be recovered.

apollo13:
Thank you for the explanation @lgfa29. Would be great if you could take the changes for a testrun and maybe merge it :)

apollo13:
@lgfa29 So when the plugin config is constructed in tests like this: `pluginConfig := PluginConfig{}` the defaults from the spec obviously don't apply. I am wondering at which point I should handle an empty driver, would it be here https://github.com/hashicorp/nomad-driver-podman/pull/285/files#diff-d16a4232641fc7ed88a3737c22b4f54d7e03dd7fdf9053530fde70e5bca63fb1R514 or should the tests create a "proper" config. Not sure enough about nomad internals to know how to proceed.

driver.go
```diff
@@ -502,20 +502,36 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
...
 		}
-	case LOG_DRIVER_JOURNALD:
-		createOpts.LogConfiguration.Driver = "journald"
-	default:
-		return nil, nil, fmt.Errorf("Invalid logging.driver option")
+		createOpts.ContainerBasicConfig.LogConfiguration.Options = driverConfig.Logging.Options
+	} else {
+		d.logger.Trace("no podman log driver provided, defaulting to plugin config")
+		switch d.config.Logging.Driver {
+		case "": // The default is nomad
```
apollo13, line 521: Pushed this as the fix for the mentioned "issue", not sure if this should go somewhere else or if it would be possible to always properly initialize the `Logging` struct.

apollo13:
@lgfa29 👋 would love a review :)

lgfa29 (approved):
Hi @apollo13, apologies for the delay here. This PR fell through the cracks during the break....

The changes look good to me! I pushed a commit to add a CHANGELOG entry and also to fix how the driver config is used.

Previously only the `Driver` value was being checked, which is documented to default to `nomad`, so users may not set this value, and I don't think we were setting it, but rather assuming an empty string to be `"nomad"`.

So if someone only set the `Options` in their task `config` they would expect `Driver` to be `nomad` and the `Options` applied to the logging config, which would not be case.

In practice we don't have any `Options` for `nomad`, so this is not likely to happen now, but just playing a bit safe here 😄 

I will run another round of the upgrade path test I mentioned tomorro just to be sure it all works as expected, but the code LGTM!

Thanks again for the contribution!

driver.go
lgfa29: Yes, I think this is the correct spot. Just changed this a little in https://github.com/hashicorp/nomad-driver-podman/pull/285/commits/79fa57034c41bbd251915e54ce5a08cfce894828#diff-d16a4232641fc7ed88a3737c22b4f54d7e03dd7fdf9053530fde70e5bca63fb1L521-R521 to avoid the use of `fallthrough` since it can be a little confusing to read.

lgfa29:
Testing went well. I had to rebase the branch to fix a merge conflict in the `README`, but I think we're good to go!

apollo13:
Thank you very much and sorry for forgetting to rebase.

lgfa29:
Ah no worries. I caused the conflict when I pushed a `CHANGELOG` entry 😅

PR #286, opened 2023-09-29
Handle errors when inspect fails.

apollo13:


shoenig (approved):
LGTM; thanks @apollo13!

PR #304, opened 2023-12-06
feat: allow set global dns servers from client plugin config for contains

ttys3:
so we can config dns for containers like:

```hcl
  config {
     dns_servers = ["192.168.8.100"]
  }
}
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -149,7 +149,7 @@ type PluginConfig struct {
 	SocketPath           string           `codec:"socket_path"`
 	ClientHttpTimeout    string           `codec:"client_http_timeout"`
 	ExtraLabels          []string         `codec:"extra_labels"`
-	DNSServers           []string     `codec:"dns_servers"`
+	DNSServers           []string         `codec:"dns_servers"`
 }
 
 // LogWarnings will emit logs about known problematic configurations
--- a/config.go
+++ b/config.go
@@ -50,6 +50,7 @@ var (
 		// task containers.  If true, logs are not forwarded to nomad.
 		"disable_log_collection": hclspec.NewAttr("disable_log_collection", "bool", false),
 		"client_http_timeout":    hclspec.NewAttr("client_http_timeout", "string", false),
+		"dns_servers":            hclspec.NewAttr("dns_servers", "list(string)", false),
 	})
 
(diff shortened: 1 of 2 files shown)


jdoss:
Why would you want to set this in the plugin section of the nomad config when you can just set DNS via https://developer.hashicorp.com/nomad/docs/job-specification/network#dns-1 in the networking block of the job?

config.go
```diff
@@ -148,6 +149,7 @@ type PluginConfig struct {
 	SocketPath           string           `codec:"socket_path"`
 	ClientHttpTimeout    string           `codec:"client_http_timeout"`
 	ExtraLabels          []string         `codec:"extra_labels"`
+	DNSServers           []string     `codec:"dns_servers"`
```
shoenig, line 152: needs to be go-fmt'd

shoenig:
The docker task driver also already has a `dns_servers` [config option](https://developer.hashicorp.com/nomad/docs/drivers/docker#dns_servers); it would make sense to make it available for podman too.

shoenig:
Also @ttys3 if you don't mind rebasing on main; there's some CI fixes in there to help getting tests working again.

jdoss:
> Do you think I am unaware of this configuration option? No, I am not. When I submitted this PR, I had already been using this solution for almost two years.

No need to get defensive. Your original PR had zero details on why this was needed and I was simply asking to better understand your use-case and the reason for the PR.  Thank you for taking the time to explain why. I appreciate it.

ttys3:
> > Do you think I am unaware of this configuration option? No, I am not. When I submitted this PR, I had already been using this solution for almost two years.
> 
> No need to get defensive. Your original PR had zero details on why this was needed and I was simply asking to better understand your use-case and the reason for the PR. Thank you for taking the time to explain why. I appreciate it.

Sorry my previous reply was a bit unkind. I take back what I said before.

ttys3:
> Also @ttys3 if you don't mind rebasing on main; there's some CI fixes in there to help getting tests working again.

@shoenig rebased and formated, PTAL

shoenig (approved):
LGTM; thanks @ttys3!

PR #321, opened 2024-02-13
Add granular control of SELinux labels for host mounts 

Juanadelacuesta:
Currently when using the podman task driver plugin if the selinuxlabel = "z" option is present in the volume configuration, all the host volume mounts will be forced to use SELinux shared volume context.

There is no way for operators to use the csi-hostpath plugin or any type of mount that is not compliant with the SELinux security, or mount a private volume. This PR introduces a new configuration option that can be set per volume mount, allowing to set different SELinux context to different volumes.

There will also be a PR on nomad to read and pass the new configuration.

Closes:

https://github.com/hashicorp/nomad-driver-podman/issues/184
https://github.com/hashicorp/nomad-driver-podman/issues/185
https://github.com/hashicorp/nomad-driver-podman/pull/208


go.mod
```diff
@@ -1,9 +1,10 @@
 module github.com/hashicorp/nomad-driver-podman
 
-go 1.21
+go 1.21.0
 
-replace (
+toolchain go1.21.1
```
shoenig, line 5: This seems a bit suspicious ... we should be able to leave the module with just `go1.21` and not force specific versions (i think?)

shoenig (approved):

Issue #331, opened 2024-03-27
Failing to Unmarshal Podman 5.0 Inspect result


hbsnmyj:
Hi folks,

It seems that podman 5.0 changed [InspectContainerConfig.Config.EntryPoint](https://github.com/containers/podman/blob/main/pkg/specgen/specgen.go#L47C13-L47C15) from string to an array, which causes an unmarshal error:

```
client.alloc_runner.task_runner: running driver failed: alloc_id= task= error="rpc error: code = Unknown desc = failed to start task, could not inspect container : json: cannot unmarshal array into Go struct field InspectContainerConfig.Config.Entrypoint of type string
```

It looks like the field is not being used right now so the easier workaround might be just remove it from the struct?

shoenig:
Yeah @hbsnmyj I think removing it would be reasonable as long as we aren't using it.

PR #332, opened 2024-03-31
Fix podman 5.0 unmarshall error.

hbsnmyj:
Podman 5.0 introduces some inspect API type changes: [commit](https://github.com/containers/podman/commit/de845a5b427b82004130f27fec5c27f397722c27), breaking the unmarshal code. Since we don't use the fields, we just delete them. See Issue #331. 

Testing:
- Manully Verified that this works against podman 4.9.3 and 5.0 (nomad 1.7.6). [lightly edited transcript](https://pastebin.com/m8g74y3J)

Note: 
- libpod has just implemented a compatible unmarshal API [here](https://github.com/containers/podman/commit/585af039d7a37580065076648550802d674dd901).

### DIFF
--- a/api/structs.go
+++ b/api/structs.go
@@ -581,7 +581,8 @@ type InspectContainerConfig struct {
 	// Container working directory
 	WorkingDir string `json:"WorkingDir"`
 	// Container entrypoint
-	Entrypoint string `json:"Entrypoint"`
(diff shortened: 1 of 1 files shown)


torrancew:
It would be great to get this merged, it appears to be breaking Archlinux Nomad+Podman 5.0 deployments

skoppe:
Time for a new release?

Issue #344, opened 2024-05-21
provide yum and apt packaging


david-yu:
Currently to install the podman driver, you would need to download the raw binary from releases.hashicorp.com. It would be great if we could use `yum` or `apt` to install the podman driver easily based on a assumed default plugin_dir like https://github.com/hashicorp/nomad-driver-exec2/pull/25/files

PR #351, opened 2024-06-04
build: create linux package during CRT build

pkazmierczak:
This PR creates .deb and .rpm Linux packages of the nomad-driver-podman
task driver. Given limitations of our packaging system the driver gets
installed in /usr/bin/ but then gets linked to /opt/nomad/plugins/ during
the install steps (and likewise removed during uninstall). This way the
plugin Just Works with the default configuration of Nomad.

Closes https://github.com/hashicorp/nomad-driver-podman/issues/344

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,7 +59,7 @@ jobs:
     strategy:
       matrix:
         goos: [linux]
-        goarch: ["arm64", "amd64"]
+        goarch: ["arm", "arm64", "amd64"]
       fail-fast: false # recommended during development
     name: Go ${{ needs.get-go-version.outputs.go-version }} ${{ matrix.goos }} ${{ matrix.goarch }} build
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,6 @@
 name: build
 
-on:
-  push:
-  workflow_dispatch:
-  workflow_call:
+on: [workflow_dispatch, push]
 
 env:
   PKG_NAME: "nomad-driver-podman"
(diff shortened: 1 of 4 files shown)


.release/linux/postrm
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+rm /opt/nomad/plugins/nomad-driver-podman
```
tgross, line 3: This will let removing the package work without error even if the user has broken the install.

```suggestion
rm -f /opt/nomad/plugins/nomad-driver-podman
```

.github/workflows/build.yml
```diff
@@ -59,32 +55,57 @@ jobs:
...
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
-        goos: ["linux"]
-        goarch: ["amd64", "arm64", "arm"]
-      fail-fast: true
-
+        goos: [linux]
+        goarch: ["arm64", "amd64"]
```
tgross, line 62: It looks like we're no longer releasing ARM 32-bit. We should add that to the changelog

tgross (approved):
LGTM!

.release/linux/postrm
shoenig: good idea; we should do this on exec2 as well

pkazmierczak: ah good catch

.github/workflows/build.yml
pkazmierczak: I could also restore `arm`? I just looked at exec2 driver to refresh the gha versions etc., didn't notice I removed `arm` accidentally. fwiw Nomad is still released on 32-bit arm so why not podman driver. what do you think?

tgross: > fwiw Nomad is still released on 32-bit arm

Oh, interesting. I thought we weren't anymore but that's 32-bit Intel. We closed https://github.com/hashicorp/nomad-enterprise/issues/1053 (internal) without ever addressing 32-bit ARM.

In the meantime, yeah it doesn't make sense to remove the build until we've removed the last 32-bit build for Nomad too.

Issue #189, opened 2022-01-29
Write logs to fifo from a driver run as non-root user


rina-spinne:
### Nomad version
Nomad v1.2.4 (55e5c49b99a6fd2bf925e7fd98d95829776c331f)

### Operating system and Environment details
Void Linux

### Issue
While trying to setup the [podman driver](https://github.com/hashicorp/nomad-driver-podman/) with a non-root user all containers fail to start due to the podman driver not being able to write on the log fifo files. The nomad client is running as root.

This happens due to both the `<alloc-dir>` and the internal `<alloc-dir>/<alloc-id>/alloc/log/.<task>.stdout.fifo` being owned by root and having only user permissions. Even if I grant the podman user the permissions to access to the `<alloc-dir>`, the fifos are being created with the [limited permissions](https://github.com/hashicorp/nomad/blob/main/client/lib/fifo/fifo_unix.go#L21).

What I can get from the [driver documentation](https://www.nomadproject.io/docs/internals/plugins/task-drivers#logging) is that drivers are expected to be able to write to the alloc logs fifo files but because of the permission design it looks like it's not possible to run any driver with a non-root user.

This works well when running podman as root.

### Reproduction steps
Start the podman service as a non-root user and try to run any job

#### Expected Result
Containers starting when running with a podman service with a non-root user

#### Actual Result
Job fails to start due to podman not being able to write to log's fifo

### Job file (if appropriate)
https://github.com/hashicorp/nomad-driver-podman/blob/main/examples/jobs/redis_deprecated.nomad

### Nomad Client logs (if appropriate)

```
$ nomad agent -dev -plugin-dir /usr/lib/nomad/plugins/ -bind 0.0.0.0
    2022-01-29T09:20:09.519Z [DEBUG] worker: dequeued evaluation: worker_id=a5f28f46-1dcb-d9d1-0ca8-da421bfe4fc3 eval_id=7c921fcf-641d-9a96-e2f1-7d955cc2fb72 type=service namespace=default job_id=redis node_id="" triggered_by=job-register
    2022-01-29T09:20:09.520Z [DEBUG] worker.service_sched: reconciled current state with desired state: eval_id=7c921fcf-641d-9a96-e2f1-7d955cc2fb72 job_id=redis namespace=default worker_id=a5f28f46-1dcb-d9d1-0ca8-da421bfe4fc3
  results=
  | Total changes: (place 1) (destructive 0) (inplace 0) (stop 0)
[...]

PR #371, opened 2024-07-24
Support for multiple podman sockets

### LINKED ISSUE #84: Run rootless containers while running nomad as root
Hi all,

I'd like to be able to run a nomad job using "rootless" style containers while nomad itself is still running as root. That is to say, I'd like to be able to run the container with user namespacing enabled, and uid 0 within the container mapped to a chosen non-root uid on the host. Nomad itself might be called upon to run "rootless" containers on behalf of multiple other uids (which removes running nomad as a non-root uid from consideration).

This Podman driver implements the {{--user}} flag the same way as docker does, all processes within the container are squashed to run as the chosen uid, and the "rootless" functionality is lost (having multiple users within the container, uid 0 having sufficient privilege to switch user, change file owners etc). This is how we currently run docker containers under nomad, but its problematic in that containers that users are developing locally using non-root podman will not behave the same way when launched under nomad, and containers from the docker hub that require processes to run as specific uids (or use multiple uids internally) fail with filesystem permission errors and similar.

Do you think it would be possible to get the "rootless" behaviour using the podman v2 http api, either by having nomad itself drop privileges to the desired user before making the api request, or by passing additional options via the api to setup the user namespacing in a way equivalent to running `podman run` as the target uid? I see that some user namespacing options exist in the code but are currently commented out (https://github.com/hashicorp/nomad-driver-podman/blob/master/api/structs.go#L297).

Thanks,
Ben

### LINKED ISSUE #284: Rootless in combination with rootful
I really like running workload in rootless containers.

However, I have not found a way to have both rootless and rootful containers on a single nomad node.

Because some things do need rootful containers, CSI plugins being one example, I see no alternative than switching to rootful.

Is there anything I am not aware about?

adriaandens:
**Problem description**: I'm using nomad with this driver to run workloads on several VPSs and I want to have support to run containers in rootless mode under different low-privilege users. The current driver only allows you to specify one rootless socket in the driver configuration. Having different host users run containers makes sure that the container UIDs don't map to the same user on the host.

**Fix**: I have added support for multiple "socket" blocks in the driver configuration. The Tasks can reference a socket using the name of the socket. As far as I have manually tested I have made it backwards compatible (but mutually exclusive) to be combined with the existing `socket_path`. I've added some code so there's always a default podman client option even if it's not specified by the user (preferably this is tackled in the HCL parsing but I didn't find how I could it handle this as an error early in the spec parsing).

Would resolve #284 (you can have a root socket and a low-priv user socket), resolve #84 (it implements exactly as the title asks)

In the driver config you can make multiple `socket` blocks mentioning the socket path, the host user and give it a name :
```hcl
  config {
    socket {
      name = "default"
      socket_path = "unix://run/user/1000/podman/podman.sock"
    }
    socket {
      name = "anotherName"
      socket_path = "unix://run/user/1001/podman/podman.sock"
    }
    # Errors with a log entry in journald if we set both the socket and the socket_path
    #socket_path = "unix://run/user/1000/podman/podman.sock"
    #disable_log_collection = true
  }
}
```
In the task HCL we can specify the socket name from the driver:
```hcl
job "deploy_registry" {
  datacenters = ["GRA"]

  group "reggie" {
    network {
      port "registry_port" {
        to = 5000
      }
      port "another_port" {
        to = 5001
      }
    }   

    task "deploy_registry_container" {
      driver = "podman"
      config {
        image = "localhost/mijnreggie:latest"
        ports = ["registry_port"]
        privileged = false
        socket = "default" # <-------------------------------------- HERE
        logging = {
          driver = "journald"
          options = [
            {
              "tag" = "reggieeeee"
            }
          ]
        }
      }
    }   

    task "deploy_registry_container_as_ansible" {
      driver = "podman"
      config {
        image = "localhost/mijnreggie:latest"
        ports = ["another_port"]
        privileged = false
        socket = "anotherName" # <-------------------------------------- AND HERE
        logging = {
          driver = "journald"
          options = [
            {
              "tag" = "reggieeeee"
            }
          ]
        }
      }
    }   
  }
}
```

The image `localhost/mijnreggie` being referenced above was built with podman from the `Dockerfile` below:
```
FROM registry:latest
RUN /usr/sbin/addgroup owla
RUN /usr/sbin/adduser -D -u 1337 -G owla owla
USER owla
```
Since podman reads containers from the local user registry, it needs to be available on each user that will run the container.

**Output of nomad running the job:**
```
$ nomad job run registry.hcl
==> 2024-07-24T19:36:52Z: Monitoring evaluation "69ce44dd"
    2024-07-24T19:36:52Z: Evaluation triggered by job "deploy_registry"
    2024-07-24T19:36:53Z: Evaluation within deployment: "6c924bcd"
    2024-07-24T19:36:53Z: Allocation "622cb564" created: node "c4616d09", group "reggie"
    2024-07-24T19:36:53Z: Evaluation status changed: "pending" -> "complete"
==> 2024-07-24T19:36:53Z: Evaluation "69ce44dd" finished with status "complete"
==> 2024-07-24T19:36:53Z: Monitoring deployment "6c924bcd"
  ✓ Deployment "6c924bcd" successful
    
    2024-07-24T19:37:06Z
    ID          = 6c924bcd
    Job ID      = deploy_registry
    Job Version = 46
    Status      = successful
    Description = Deployment completed successfully
    
    Deployed
    Task Group  Desired  Placed  Healthy  Unhealthy  Progress Deadline
    reggie      1        1       1        0          2024-07-24T19:47:04Z
```

The processes as seen with `ps`:
```
# ps -ef | grep "registry serve"
101336    271845  271843  0 19:36 ?        00:00:00 registry serve /etc/docker/registry/config.yml
166872    271868  271866  0 19:36 ?        00:00:00 registry serve /etc/docker/registry/config.yml
```

`/etc/subuid` maps the two container users to subordinate users on the host (so the UID above is begin of range + uid 1337 - 1):
```
# cat /etc/subuid
debian:100000:65536
ansible:165536:65536
...
```
You can also inspect the containers with `podman ps` under the user running the container:
```
# su - debian
$ podman ps
CONTAINER ID  IMAGE                        COMMAND               CREATED        STATUS            PORTS                                                       NAMES
336f1a262f77  localhost/mijnreggie:latest  /etc/docker/regis...  6 minutes ago  Up 6 minutes ago  x.x.x.x:29351->5000/tcp, x.x.x.x:29351->5000/udp  deploy_registry_container-622cb564-624c-ec67-4b19-e2452459406c
```

**To do**:
* [x] Gather feedback on whether this request is mergeable in the future? Or whether it conflicts with the roadmap of this plugin. 
* [x] Verify backward compatibility
* [x] Implement further tests
* [x] Squat more bugs
* [x] Update documentation
* [ ] Path to deprecate previous single `socket_path` implementation
* [ ] Other feedback

**Feedback needed around using the "user" field**: The user field in Task HCL is currently used to decide what user is run inside the container but this would better used to specify the host user that runs the container task. An additional parameter `container_user` could be added to give to `podman` commands to specify which user to run the `CMD` as inside the container. I understood this is for compatibility with the docker driver but it's confusing. If wanted, I can add this implementation to this PR too but it'll break backwards compatibility. 

**Not in scope of this PR**: Fixing the open issue with FIFO logging #189 . This to me looks like a problem in nomad itself that creates a FIFO without looking at the task "user" field to set the owner correctly ([newLogRotatorMapper](https://github.com/hashicorp/nomad/blob/92d216f3b8c4945dd6ab8056cd2574768dfb8c98/client/logmon/logmon.go#L208) calls the mkfifo [here](https://github.com/hashicorp/nomad/blob/92d216f3b8c4945dd6ab8056cd2574768dfb8c98/client/lib/fifo/fifo_unix.go#L24) ). When specifying `journald` the driver does not crash because it avoids the FIFO issues. If we remap the `user` key to be the host user, I can make a PR to fix the FIFO problem (since then we know the podman user and can change the rights on the pipe)

Any feedback is welcome.


adriaandens:
@tgross Do you have any feedback on this PR? Or time in the coming weeks to review this? It's not urgent but I don't want it to die a silent dead either.

skoppe:
Awesome stuff, and would remove an ugly workaround we have for running both rootful and rootless podman containers on the same node.

api/container_delete.go
```diff
@@ -20,7 +20,7 @@ func (c *API) ContainerDelete(ctx context.Context, name string, force bool, dele
 
 	defer ignoreClose(res.Body)
 
-	if res.StatusCode == http.StatusNoContent {
+	if res.StatusCode == http.StatusNoContent || res.StatusCode == http.StatusOK {
```
tgross, line 23: Is this fixing a bug you saw or is this somehow changed for this feature?

driver_test.go
```diff
@@ -1955,12 +1960,18 @@ func TestPodmanDriver_SignalTask(t *testing.T) {
 func TestPodmanDriver_Sysctl(t *testing.T) {
 	ci.Parallel(t)
 
+	// Test fails if not root
+	if os.Getuid() != 0 {
```
tgross, line 1964: Should this be [os.Geteuid](https://pkg.go.dev/os#Geteuid) or does it matter?

driver_test.go
```diff
@@ -2035,7 +2046,7 @@ func startDestroyInspectImage(t *testing.T, image string, taskName string) {
 		AllocID:   uuid.Generate(),
 		Resources: createBasicResources(),
 	}
-	imageID, err := getPodmanDriver(t, d).createImage(image, &TaskAuthConfig{}, false, false, 5*time.Minute, task)
+	imageID, err := getPodmanDriver(t, d).createImage(image, &TaskAuthConfig{}, false, false, getPodmanDriver(t, d).defaultPodman, 5*time.Minute, task)
```
tgross, line 2049: Nitpicky, but maybe something like:

```suggestion
	drv := getPodmanDriver(t, d)
	imageID, err := drv.createImage(image, &TaskAuthConfig{}, false, false, drv.defaultPodman, 5*time.Minute, task)
```

Similar for a lot of these below.

driver.go
```diff
@@ -325,23 +386,35 @@ func (d *Driver) RecoverTask(handle *drivers.TaskHandle) error {
...
 	h := &TaskHandle{
 		containerID:           taskState.ContainerID,
 		driver:                d,
+		podmanClient:          taskPodmanClient,
 		taskConfig:            taskState.TaskConfig,
 		procState:             drivers.TaskStateUnknown,
 		startedAt:             taskState.StartedAt,
 		exitResult:            &drivers.ExitResult{},
-		logger:                d.logger.Named("podmanHandle"),
+		logger:                d.logger.Named("podmanHandle"), // TODO: does this need to be podmanClient aware?
```
tgross, line 417: It'd be nice!

driver.go
```diff
@@ -293,6 +353,7 @@ func (d *Driver) buildFingerprint() *drivers.Fingerprint {
 		d.cgroupMgr = info.Host.CgroupManager
 	}
 
+	// TODO: Can we make this more correct with one or the other podman client?
```
tgross, line 356: Yeah for the `driver.podman.version`, `driver.podman.rootless`, and `driver.podman.cgroupVersion`, you'll need to create an attribute for the non-default config and fingerprint each one separately. You can't just interpolate "default" in there, because that'll break existing constraints. See https://github.com/hashicorp/nomad/blob/v1.9.0-beta.1/client/fingerprint/vault.go#L96-L106 for an example of where this kind of thing was added in Nomad for multiple Vault configs.

driver_test.go
```diff
@@ -1043,7 +1048,7 @@ func TestPodmanDriver_Init(t *testing.T) {
 
 	// podman maps init process to /dev/init, so we should see this
```
tgross, line 1049: You've changed the assertion but not the comment. It's not obvious to me why the assertion changed though.

driver.go
```diff
@@ -325,23 +386,35 @@ func (d *Driver) RecoverTask(handle *drivers.TaskHandle) error {
...
+			podmanClientName = name
+			taskPodmanClient = podmanClient
+			break
+		} else if errors.Is(err, api.ContainerNotFound) {
+			d.logger.Debug("Recovery lookup found no container", "task", handle.Config.ID, "container", taskState.ContainerID, "with podman client", name, "error", err)
+		} else if err != nil {
+			d.logger.Warn("Recovery lookup failed", "task", handle.Config.ID, "container", taskState.ContainerID, "with podman client", name, "error", err)
+			// TODO: Maybe not return here?
+			return nil
+		}
```
tgross, line 406: You need to handle the case where it wasn't found for any client.

driver.go
```diff
@@ -325,23 +386,35 @@ func (d *Driver) RecoverTask(handle *drivers.TaskHandle) error {
...
+		if err == nil {
+			podmanClientName = name
+			taskPodmanClient = podmanClient
+			break
+		} else if errors.Is(err, api.ContainerNotFound) {
+			d.logger.Debug("Recovery lookup found no container", "task", handle.Config.ID, "container", taskState.ContainerID, "with podman client", name, "error", err)
+		} else if err != nil {
+			d.logger.Warn("Recovery lookup failed", "task", handle.Config.ID, "container", taskState.ContainerID, "with podman client", name, "error", err)
+			// TODO: Maybe not return here?
+			return nil
```
tgross, line 405: If you're getting an error other than `ContainerNotFound` there's a communications issue. We could return the error here or clear `err` if the container is found later.

driver.go
```diff
@@ -325,23 +386,35 @@ func (d *Driver) RecoverTask(handle *drivers.TaskHandle) error {
...
-		return nil
-	} else if err != nil {
-		d.logger.Warn("Recovery lookup failed", "task", handle.Config.ID, "container", taskState.ContainerID, "error", err)
-		return nil
+	// TODO: It can be created by any podman client, so we need to iterate over all podman clients and find that container back...
+	var podmanClientName string
+	var taskPodmanClient *api.API
+	var inspectData api.InspectContainerData
+	var err error
+	for name, podmanClient := range d.podmanClients {
```
tgross, line 394: When we recover a task, don't we have the task config handy and so we should be able to use that to figure out which client to use, rather than iterating?

driver.go
```diff
@@ -196,22 +196,81 @@ func (d *Driver) SetConfig(cfg *base.Config) error {
...
+			d.podmanClients["default"] = d.newPodmanClient(timeout, fmt.Sprintf("unix:///run/user/%d/podman/podman.sock", uid), "", true)
+		}
+		d.defaultPodman = d.podmanClients["default"]
+	}
 	d.compute = cfg.AgentConfig.Compute()
 
 	return nil
 }
 
+func (d *Driver) makePodmanClients(sockets []PluginSocketConfig, timeout time.Duration) map[string]*api.API {
```
tgross, line 222: In various parts of Nomad core, we've handled this "default config block" problem by requiring the _name_ of the default config be "default" and interpolate the `name = ""` as `name = "default"`. That reduces the amount of logic you need here and lets users omit the name field if they have only a single default config.

tgross (changes requested):
Hi @adriaandens! I've left some comments around configuration especially that I think could use a little redesign, but overall this is looking really good!

driver_test.go
adriaandens: The container init binary on my system was mounted on /run/podman-init, as described on [the manpage of podman-run](https://docs.podman.io/en/latest/markdown/podman-run.1.html).

api/container_delete.go
adriaandens: I had to make this change to make a test pass. I'll have to check which one was failing to see if it was a bug. But it's not particularly related to the multiple podman socket feature.

driver.go
adriaandens: I've changed the implementation as described. It would've been nicer if I was able to push some of the logic to the HCL config parser (for example demanding no duplicate names, limiting the characters that are allowed, ...)

driver.go
adriaandens: Yup, I figured it out. In `StartTask()` I found that you can decode the TaskHandle with `DecodeDriverConfig()` to get a handle on a more specific struct that contains all the task fields (incl. the socket name). I did rename this struct to `podmanTaskConfig` in StartTask (and RecoverTask) because it's more clear than `driverConfig`.

driver.go
adriaandens: I've added a whole bunch of attributes that are useful to expose, all in the form `driver.podman.<socket name>.<attr>`. This means there's no `driver.podman.version`, `driver.podman.rootless` or `driver.podman.cgroupVersion` anymore. It also doesn't make sense to have a `driver.podman.rootless` since you can combine rootless/root sockets and the value changes per socket.

Screenshot_2024-09-29_16-26-39

driver.go
adriaandens: I've changed it to "podman.<socket name>" but I don't see it reflected in journald/journalctl output?

driver_test.go
adriaandens: Updated

driver_test.go
adriaandens: I'm not sure if there's a case where the tests will change their UID so that EUID would be different from UID call. But it doesn't hurt to change it to Geteuid.

api/api.go
```diff
@@ -18,34 +18,40 @@ import (
...
 	}
 	uid := os.Getuid()
-	// are we root?
 	if uid == 0 {
 		cfg.SocketPath = "unix:///run/podman/podman.sock"
 	} else {
-		// not? then let's try the default per-user socket location
 		cfg.SocketPath = fmt.Sprintf("unix:///run/user/%d/podman/podman.sock", uid)
 	}
+	cfg.HostUser = "root"
```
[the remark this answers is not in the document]
adriaandens: I removed the HostUser field from the config block, since it was not used anywhere in the driver. There were also places in the `driver.go` code where I couldn't do better than just setting an empty string. Removing the host user, removes the issue raised :)

driver.go
adriaandens: There's also a change to the driver struct and how it gives back "healthy" or "unhealthy". Since there's multiple podman sockets, it reports "healthy" if it can connect and retrieve info from all of them. Unhealthy if it cannot connect to some of the sockets and undetected if it cannot connect to any. I've also added a sort on the unresponsive sockets because it'll keep generating logs otherwise: 
Screenshot_2024-09-29_16-47-52

tgross (changes requested):
@adriaandens there's a few items the linter just picked up and a couple of as-yet-unresolved comments from before. Another minor item I didn't notice before is that you made all the source files executable (i.e. changed from 0644 to 0755). Can you fix that too?

> If you're OK with the new HCL config implementation (a socket path and name (if omitted, resolves to "default")), I can update the README, docs, and also write some examples so it's easy for people to switch over their configs

Sounds good!  We'll want a changelog entry as well.

api/container_delete.go
adriaandens: If I run `go test -v -race` without this extra check I get the following output/errors in all tasks:
```
2024-10-01T18:57:17.643+0100 [DEBUG] nomad-driver-podman/handle.go:79: podman.podman.default: Starting exitWatcher: container=328b6c260830fad1a9d87a05e7e4f480ca80af95eef71f33a0f600860a39d809
2024-10-01T18:57:17.643+0100 [DEBUG] nomad-driver-podman/handle.go:216: podman.podman.default: Container is not running anymore: container=328b6c260830fad1a9d87a05e7e4f480ca80af95eef71f33a0f600860a39d809 error="No such Container"
2024-10-01T18:57:18.644+0100 [DEBUG] nomad-driver-podman/handle.go:82: podman.podman.default: Stopping exitWatcher: container=328b6c260830fad1a9d87a05e7e4f480ca80af95eef71f33a0f600860a39d809
2024-10-01T18:57:18.724+0100 [WARN]  nomad-driver-podman/driver.go:1272: podman: Could not remove container: container=328b6c260830fad1a9d87a05e7e4f480ca80af95eef71f33a0f600860a39d809 error="cannot delete container, status code: 200"
```
It states "cannot delete container, status code: 200". The output from Podman is the following JSON:
```
[{"Id":"328b6c260830fad1a9d87a05e7e4f480ca80af95eef71f33a0f600860a39d809","RawInput":"328b6c260830fad1a9d87a05e7e4f480ca80af95eef71f33a0f600860a39d809"}]
```
Adding the extra check makes sure that if Podman gives a "200 OK" back, that we don't log this as an error.

driver.go
adriaandens: Since I'm not looping over all podman clients anymore, if we receive an error that's not ContainerNotFound, we return the error.

README.md
```diff
@@ -148,7 +148,7 @@ plugin "nomad-driver-podman" {
 }
 ```
 
-* socket_path (string) Defaults to `"unix:///run/podman/podman.sock"` when running as root or a cgroup V1 system, and `"unix:///run/user/<USER_ID>/podman/podman.sock"` for rootless cgroup V2 systems
+* socket_path (string) Defaults to `"unix:///run/podman/podman.sock"` when running as root or a cgroup V1 system, and `"unix:///run/user/<USER_ID>/podman/podman.sock"` for rootless cgroup V2 systems. Mutually exclusive with socket_path option.
```
tgross, line 151:
```suggestion
* socket_path (string) Defaults to `"unix:///run/podman/podman.sock"` when running as root or a cgroup V1 system, and `"unix:///run/user/<USER_ID>/podman/podman.sock"` for rootless cgroup V2 systems. Mutually exclusive with `socket` block.
```

README.md
```diff
@@ -158,6 +158,26 @@ plugin "nomad-driver-podman" {
 }
 ```
 
+* socket stanza: If you require to use multiple podman sockets (rootful/rootless), you can define a list of them. Mutually exclusive with socket_path option.
```
tgross, line 161: HCL has "blocks", not "stanzas" (lots of docs have this wrong, unfortunately). Also, because the `socket` block itself has a `socket_path`, it'd be good to disambiguate which field we mean just to be super-clear. How about:

```suggestion
* socket block: Configures a single podman socket. You can define multiple `socket` blocks if you need to use multiple podman sockets (for example, rootless vs rootful sockets). Mutually exclusive with the top-level `plugin.config.socket_path` option.
```

README.md
```diff
@@ -403,6 +423,14 @@ config {
 }
 ```
 
+* **socket** - (Optional) The name of the socket as defined in the socket stanza in `client.hcl`. Defaults to "default" as socket.
```
tgross, line 426:
```suggestion
* **socket** - (Optional) The name of the socket as defined in the socket block in the client agent's plugin configuration. Defaults to the socket named "default".
```

tgross:
Sorry @adriaandens, the new HCL files need `hclfmt`'ing so the tests run, too.

driver.go
tgross: > It would've been nicer if I was able to push some of the logic to the HCL config parser

Yeah, we don't really provide a good "driver SDK" for that kind of thing.

tgross (changes requested):

tgross:
> I ran into the issue of the /dev/init vs. /run/podman-init because I'm running Podman 4.3.1 (Debian stable podman package on bookworm) whilst the Github Action uses 3.4.4. From going through the versions on the Podman docs, the /run/podman-init is mentioned from v4.2 onwards.

Yeah this really isn't part of the API _per se_ but our test is making assertions on internal implementation details that we probably shouldn't be. Out of scope for this PR to worry about.

> For the Github Actions, it would (although outside of scope for this PR) make sense to run all the tests with multiple Podman versions: the minimal supported by this driver, latest, stable, latest from each major version (3.4.4, 4.9.3, 5.2.3), ... so that there's coverage of the most used versions.

Probably a good idea, but yeah outside of the scope of this PR for sure.

I think we're getting close here, but `TestPodmanDriver_Pull_Timeout` is currently failing.

adriaandens:
> I think we're getting close here, but `TestPodmanDriver_Pull_Timeout` is currently failing.

My PR removes the slowPodman driver (a driver with no http timeout) since having multiple podman sockets would mean additional slowPodman drivers for each. I fixed the issue now by using the streaming http client in the existing driver implementation if a context is passed with a deadline higher than the timeout of the http client in the api.

tgross (approved):
LGTM :+1: 

Spending a bit of time looking at `StartTask` is making me think there are some investments we could make in refactoring here. And as noted in the discussion, we should follow-up with improving the test matrix to cover versions of podman we intend to support (it's not clear to me what we've decided there). But this is good-to-merge as-is.

PR #386, opened 2024-10-08
(Regression) Tests and bugfixes for multiple socket implementation

adriaandens:
This PR contains some fixes in the podman driver behaviour when there are multiple sockets in use. Additionally, I've written regression tests that test the behaviour of having multiple sockets.

Fixes:
* If there was one socket with a name different from "default", the attributes would not get prefixed with `driver.podman.attrName` but with `driver.podman.<name of socket>.attrName`. Since this would break existing nomad scheduling, now if a socket is the default socket (`isDefaultClient()` returns true), attributes are written with the `driver.podman.attrName` syntax. Also added `driver.podman.[<name of socket>.]socketName attribute so it's still possible to find the name when it's the default (+ makes it easier if you're using attribute filtering to verify podman has the low-priv socket you want)
fix1
* I implemented a clean string function `cleanUpSocketName()` for the purpose of having nice attribute names but by cleaning it before using it as a key in podmanClient map, it made it so that it was no longer "found" by other operations using the map and not cleaning the socket name. Since the internal representation of the socket name as a key in a map is safe, I removed the cleanup function and only apply it where I wanted it: in the attributes.

`driver_test.go` contains the Tests that validate this behaviour.

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -222,7 +222,7 @@ func (d *Driver) makePodmanClients(sockets []PluginSocketConfig, timeout time.Du
 		if sock.Name == "" {
 			sock.Name = "default"
 		}
-		sock.Name = cleanUpSocketName(sock.Name)
+		sock.Name = sock.Name
 		if sock.Name == "default" && !foundDefaultPodman {
 			foundDefaultPodman = true
 			podmanClient = d.newPodmanClient(timeout, sock.SocketPath, true)
@@ -270,7 +270,7 @@ func (d *Driver) getPodmanClient(clientName string) (*api.API, error) {
 	if ok {
 		return p, nil
 	}
-	return nil, fmt.Errorf("podman client with name %s was not found, check your podman driver config", clientName)
+	return nil, fmt.Errorf("podman client with name '%s' was not found, check your podman driver config", clientName)
 }
 
 // newPodmanClient returns Podman client configured with the provided timeout.
@@ -332,8 +332,8 @@ func (d *Driver) buildFingerprint() *drivers.Fingerprint {
 	for name, podmanClient := range d.podmanClients {
 		// Ping podman api
 		apiVersion, err := podmanClient.Ping(d.ctx)
-		attrPrefix := fmt.Sprintf("driver.podman.%s", name)
-		if name == "default" {
+		attrPrefix := fmt.Sprintf("driver.podman.%s", cleanUpSocketName(name))
+		if name == "default" || podmanClient.IsDefaultClient() {
 			attrPrefix = "driver.podman"
 		}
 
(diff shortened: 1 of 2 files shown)


driver.go
```diff
@@ -270,7 +269,7 @@ func (d *Driver) getPodmanClient(clientName string) (*api.API, error) {
 	if ok {
 		return p, nil
 	}
-	return nil, fmt.Errorf("podman client with name %s was not found, check your podman driver config", clientName)
+	return nil, fmt.Errorf("podman client with name '%s' was not found, check your podman driver config", clientName)
```
tgross, line 272:
```suggestion
	return nil, fmt.Errorf("podman client with name %q was not found, check your podman driver config", clientName)
```

driver.go
```diff
@@ -535,7 +535,7 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 		var err error
 		podmanClient, err = d.getPodmanClient(podmanTaskSocketName)
 		if err != nil {
-			return nil, nil, fmt.Errorf("podman client with name %s not found, check your podman driver config", podmanTaskSocketName)
+			return nil, nil, fmt.Errorf("podman client with name '%s' not found, check your podman driver config", podmanTaskSocketName)
```
tgross, line 538:
```suggestion
			return nil, nil, fmt.Errorf("podman client with name %q not found, check your podman driver config", podmanTaskSocketName)
```

driver.go
```diff
@@ -332,8 +331,8 @@ func (d *Driver) buildFingerprint() *drivers.Fingerprint {
 	for name, podmanClient := range d.podmanClients {
 		// Ping podman api
 		apiVersion, err := podmanClient.Ping(d.ctx)
-		attrPrefix := fmt.Sprintf("driver.podman.%s", name)
-		if name == "default" {
+		attrPrefix := fmt.Sprintf("driver.podman.%s", cleanUpSocketName(name))
+		if name == "default" || podmanClient.IsDefaultClient() {
```
tgross, line 335: When we talked thru #371, the configuration we landed on was that an omitted name or the name "default" became the default client. Given that, the bug here is that we're not accounting for the `name == ""`, right? If we're setting the default client field correctly at config parsing time, can we just make this entire conditional like this?

```suggestion
		if podmanClient.IsDefaultClient() {
```

adriaandens: I had to check the logic in `makePodmanClients()` but indeed this is the case. A socket without a name or with name default becomes the default socket. If none is found between all sockets, the first one becomes the default client. So checking `IsDefaultClient()` is sufficient as a condition.

tgross (approved):
LGTM

PR #382, opened 2024-09-16
config: allow setting `security_opt` option

ruspaul013:
Support --security-opt option as was mention in #297.

PR #397, opened 2024-12-13
Fix error variable format

ruspaul013:
Fix error variable format introduced in #382


ruspaul013:
@shoenig I made the change

shoenig (approved):

Issue #412, opened 2025-02-24
Driver of version 0.6.2 fails to start


kolotaev:
Hello 👋 
I'm starting nomad service with the version of `0.6.2` of the `nomad-driver-podman` and the plugin fails to be detected.

The trace logs I got are:

```
2025-02-25T00:40:08.478+0400 [DEBUG] client.driver_mgr: using plugin: driver=podman version=2
2025-02-25T00:40:08.478+0400 [TRACE] client.driver_mgr.stdio: waiting for stdio data: driver=podman
2025-02-25T00:40:08.479+0400 [DEBUG] client.driver_mgr.nomad-driver-podman: http baseurl: driver=podman @module=podman url=unix:///run/podman/podman.sock timestamp="2025-02-25T00:40:08.479+0400"
2025-02-25T00:40:08.534+0400 [DEBUG] client.driver_mgr.nomad-driver-podman: panic: runtime error: invalid memory address or nil pointer dereference: driver=podman
2025-02-25T00:40:08.534+0400 [DEBUG] client.driver_mgr.nomad-driver-podman: [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x9609fc]: driver=podman
2025-02-25T00:40:08.534+0400 [DEBUG] client.driver_mgr.nomad-driver-podman: driver=podman
2025-02-25T00:40:08.534+0400 [DEBUG] client.driver_mgr.nomad-driver-podman: goroutine 26 [running]:: driver=podman
2025-02-25T00:40:08.534+0400 [DEBUG] client.driver_mgr.nomad-driver-podman: main.(*Driver).buildFingerprint(0x40003bad80): driver=podman
2025-02-25T00:40:08.534+0400 [DEBUG] client.driver_mgr.nomad-driver-podman: 	github.com/hashicorp/nomad-driver-podman/driver.go:373 +0xeac: driver=podman
2025-02-25T00:40:08.534+0400 [DEBUG] client.driver_mgr.nomad-driver-podman: main.(*Driver).handleFingerprint(0x40003bad80, {0xd793c8, 0x4000010e70}, 0x400043ad90): driver=podman
2025-02-25T00:40:08.534+0400 [DEBUG] client.driver_mgr.nomad-driver-podman: 	github.com/hashicorp/nomad-driver-podman/driver.go:320 +0x80: driver=podman
[...]

PR #415, opened 2025-02-26
fingerprint: handle nil pointers in SystemInfo client response

mismithhisler:
We should always check for nil pointers when accessing response data from the podman client to eliminate the possibly of a panic due to a nil pointer dereference.

Fixes [GH #412](https://github.com/hashicorp/nomad-driver-podman/issues/412)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+BUG FIXES:
+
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM :+1: 

Let's add a changelog entry, too.

jrasell (approved):
LGTM, thanks @mismithhisler!

PR #424, opened 2025-03-25
Fix: use `volumes` instead of `v` for libpod API to delete container

ruspaul013:
Based on the discussion from https://github.com/containers/podman/discussions/25657, the correct parameter for volume is `volumes`.

Using `v`, will not remove the anonymous volumes created.

### DIFF
--- a/api/container_delete.go
+++ b/api/container_delete.go
@@ -13,7 +13,7 @@ import (
 // It takes the name or ID of a container.
 func (c *API) ContainerDelete(ctx context.Context, name string, force bool, deleteVolumes bool) error {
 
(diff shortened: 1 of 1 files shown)


Juanadelacuesta:
Hi @ruspaul013 Thank you for submitting this PR, it looks like you are correct and the [code](https://github.com/containers/podman/blob/e65687291a1b59f98d6e41477f15171a384f93a0/pkg/api/handlers/compat/containers.go#L45-L70) requieres `volumes` instead of `v`. Im going to go ahead and approve this, great job!

Juanadelacuesta (approved):

PR #439, opened 2025-05-06
fix tmpfs parser

### LINKED ISSUE #290: tmpfs options are not parsed
According to the docs, I should be able to specify some options when creating a tmpfs mount - https://docs.podman.io/en/latest/markdown/podman-run.1.html#tmpfs-fs

I've specified these in the nomad job file as:

```
        tmpfs = [
          "/tmp/csi/keys:rw,size=1000k"
        ]
```

But it seems not to be working, when I `podman container inspect` directly I get:

```
...
               "Tmpfs": {
                    "/tmp/csi/keys:rw,size=1000k": "rw,rprivate,nosuid,nodev,tmpcopyup"
               },
...
```

Which rather suggests that it's mounted a folder named `/tmp/csi/keys:rw,size=1000k` rather than `/tmp/csi/keys`

* nomad-driver-podman 0.5.1
* podman 4.3.1

I'm very new to this, so maybe I'm missing something obvious...

ruspaul013:
Fixes #290

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -1524,8 +1524,10 @@ func (d *Driver) containerMounts(task *drivers.TaskConfig, driverConfig *TaskCon
 	}
 
 	for _, dst := range driverConfig.Tmpfs {
+		splitTmpfs := strings.Split(dst, ":")
 		bind := spec.Mount{
-			Destination: dst,
+			Destination: splitTmpfs[0],
+			Options:     strings.Split(splitTmpfs[1], ","),
 			Type:        "tmpfs",
 		}
 		binds = append(binds, bind)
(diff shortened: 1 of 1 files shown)


driver.go
```diff
@@ -1524,8 +1524,10 @@ func (d *Driver) containerMounts(task *drivers.TaskConfig, driverConfig *TaskCon
 	}
 
 	for _, dst := range driverConfig.Tmpfs {
+		splitTmpfs := strings.Split(dst, ":")
 		bind := spec.Mount{
-			Destination: dst,
+			Destination: splitTmpfs[0],
+			Options:     strings.Split(splitTmpfs[1], ","),
```
mismithhisler, line 1530: hmm I'm not sure we are guaranteed to have these options, so splitting on a `:` may only result in a slice length 1.  Can we do a check somewhere around here to avoid any array indexing errors?

ruspaul013: sure, thanks for suggestion

mismithhisler: Thank you for the contribution!

mismithhisler (approved):
LGTM, thanks!

Issue #401, opened 2024-12-16
cannot unmarshal InspectNetworkSettings.NetworkSettings.SecondaryIPAddresses of type string"


victorserbu2709:
I encountered following error while trying to create a patch for https://github.com/hashicorp/nomad/issues/11085
```
   2024-12-16T22:24:41.186+0200 [ERROR] client.driver_mgr.nomad-driver-podman: failed to inspect container: driver=podman @module=podman error="json: cannot unmarshal object into Go struct field InspectNetworkSettings.NetworkSettings.SecondaryIPAddresses of type string" timestamp="2024-12-16T22:24:41.185+0200"
    2024-12-16T22:24:41.186+0200 [DEBUG] client.driver_mgr.nomad-driver-podman: Cleaning up: driver=podman @module=podman container=2274a57f2913499b6ac48cfe6ca56e5418aa8b30e21790827e8bec731fb830da timestamp="2024-12-16T22:24:41.185+0200"
    2024-12-16T22:24:41.589+0200 [DEBUG] nomad: memberlist: Stream connection from=127.0.0.1:57566
    2024-12-16T22:24:44.847+0200 [DEBUG] http: request complete: method=GET path=/v1/agent/health?type=client duration="216.508µs"
    2024-12-16T22:24:49.292+0200 [DEBUG] http: request complete: method=GET path=/v1/agent/health?type=server duration=1.193728ms
    2024-12-16T22:24:51.600+0200 [DEBUG] nomad: memberlist: Stream connection from=127.0.0.1:42830
```

If I do podman inspect on a container with multiple network interface I obtain
```
...
          "NetworkSettings": {
               "EndpointID": "",
               "Gateway": "172.26.64.1",
               "IPAddress": "172.26.64.36",
               "IPPrefixLen": 20,
               "SecondaryIPAddresses": [
                    {
                         "Addr": "172.27.64.31",
                         "PrefixLength": 24
                    }
               ],
               "IPv6Gateway": "",
               "GlobalIPv6Address": "",
               "GlobalIPv6PrefixLen": 0,
               "MacAddress": "72:a4:c6:5c:3f:e8",
               "AdditionalMACAddresses": [
                    "72:b9:01:bf:9f:e1"
               ],
               "Bridge": "",
...
```

PR #442, opened 2025-06-02
Added Address type

TheDevtop:
This type enables the plugin to properly handle secondary IP addresses.
As the current Podman API requires [ref](https://docs.podman.io/en/latest/_static/api.html#tag/containers/operation/ContainerInspectLibpod).

### Origins

The DHCP server in my network prioritizes IPv6 over IPv4. When starting a podman container via the CNI (macvlan+dhcp) it would error `json: cannot unmarshal object into Go struct field Inspect Network Settings.Network Settings.S econdaryIPv6Addresses of type string`.

After some research I found out that Podman indeed returns a list of `Address` instead of a single string.

### DIFF
--- a/api/structs.go
+++ b/api/structs.go
@@ -519,6 +519,13 @@ type PortMapping struct {
 	Protocol string `json:"protocol,omitempty"`
 }
 
+// IP address block
+// Needed since Podman 5.0
+type Address struct {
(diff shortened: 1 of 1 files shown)


TheDevtop:
Related to #401

PR #448, opened 2025-06-30
Fixes #444 Fix softlink deletion on rpm upgrades

### LINKED ISSUE #444: RHEL9 install broken.
The [official installation instructions](https://developer.hashicorp.com/nomad/plugins/drivers/podman) do not work. 
After running `sudo yum -y install nomad-driver-podman` and restarting both Podman and Nomad, Nomad does not find the Podman driver.

Manually downloading the current dev release from Github and placing it in the plugins directory also does not work.

Klaas-:
See https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_syntax and https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#ordering for details, have not tested it, can't manually build your packages but should work

### DIFF
--- a/.release/linux/postrm
+++ b/.release/linux/postrm
@@ -1,4 +1,6 @@
 #!/bin/sh
 
-# Remove symlink from plugins directory.
-rm -f /opt/nomad/data/plugins/nomad-driver-podman
+if [ $1 -eq 0 ] ; then
+  # Remove symlink from plugins directory on uninstall
+  rm -f /opt/nomad/data/plugins/nomad-driver-podman
+fi


Klaas-:
If you want to improve it even further, don't define the symlink in the scripts, but make it part of your actual rpm. That way also rpm -qf /opt/nomad/data/plugins/nomad-driver-podman will show it's owned by your package

Juanadelacuesta:
hell @Klaas- Thank you for contributing to the nomad project!
For legal reasons we can not accept it without the CLA agreement, is there any why you can sign it?

Klaas-:
> hell @Klaas- Thank you for contributing to the nomad project! For legal reasons we can not accept it without the CLA agreement, is there any why you can sign it?

I can't do that, not working in my browser, so you'll have to either do those changes yourself (or even better do not make a symlink in scripts but rather in the rpm.

I can however say the changes are trivial and I put them here under "Mozilla Public License, version 2.0" like the whole repo is licensed :D

PR #454, opened 2025-07-23
use ContainerKill instead of ContainerStop to StopTask

dylrich:
I wanted to start a conversation about signal handling -- I noticed that the Docker driver was sending a kill signal to the container instead of the stop command and thought it would be nice if the Podman driver did this as well. I wasn't able to run all of the tests successfully locally on clone, so I'm unsure if this breaks some other expectation.

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -1303,7 +1303,7 @@ func (d *Driver) StopTask(taskID string, timeout time.Duration, signal string) e
 	}
 
 	// fixme send proper signal to container
-	err := handle.podmanClient.ContainerStop(d.ctx, handle.containerID, int(timeout.Seconds()), true)
+	err := handle.podmanClient.ContainerKill(d.ctx, handle.containerID, signal)
 	switch {
 	case err == nil:
 		return nil


adriaandens:
[Container stop](https://docs.podman.io/en/latest/markdown/podman-stop.1.html) sends the SIGTERM and SIGKILL signals after the timeout has been reached. Setting the timeout to 0 is the same as killing, if that's what you're after?

tgross (changes requested):
Thanks for the PR @dylrich! 

This would allow the [`task.kill_signal`](https://developer.hashicorp.com/nomad/docs/job-specification/task#kill_signal) to be specified, which is great. But then we need to follow-up with sending `SIGKILL` when the [`task.kill_timeout`](https://developer.hashicorp.com/nomad/docs/job-specification/task#kill_timeout) expires. 

Because the Podman API can't do that for us with `ContainerKill`, you'll need to implement that here as well. As you've noticed in your comment, that'll be similar to what we've done for the in-tree `docker` driver [here](https://github.com/hashicorp/nomad/blob/26554e544e2660b06d4153ddb64cd21688e851ed/drivers/docker/handle.go#L164).

PR #476, opened 2025-12-02
Add support for pasta networking mode

### LINKED ISSUE #345: Podman pasta usermode networking
I am looking to implement pasta usermode networking for rootless containers. Pasta is a replacement of `slirp4netns`.

See https://docs.podman.io/en/latest/markdown/podman-run.1.html#network-mode-net (scroll down to `pasta[:OPTIONS,…]`).

Aside from being (supposedly) faster than `slirp4netns`, it should support inter-container networking better. Allowing things like sidecars and Nomad inter task communication.

Before all that, I am trying to gauge support. So, is pasta rootless networking in line with the project?

In terms of implementation, even though it is the default in Podman > 5, I am thinking of requiring explicit boolean in plugin config to toggle it on. It also requires the pasta binary to be present in podman's info output.

sigmaris:
Adds support for `pasta` networking mode, which is another user space networking mode, similar to `slirp4netns`. When using the Podman CLI tools, it replaces `slirp4netns` as the default mode for rootless containers in podman >= 5.0.0, and I've adopted the same default behaviour (if no network mode is specified for a task) here.

Closes #345

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## UNRELEASED
 
+* api: Support pasta network_mode and use it as default for rootless containers on Podman 5.0+ [[GH-476](https://github.com/hashicorp/nomad-driver-podman/pull/476)]
 * build: Update Nomad version to 1.11.0 [[GH-478](https://github.com/hashicorp/nomad-driver-podman/pull/478)]
 * build: Updated to Go 1.25.5 [[GH-477](https://github.com/hashicorp/nomad-driver-podman/pull/477)]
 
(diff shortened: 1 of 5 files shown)


tgross (commented):
LGTM! I've been able to verify this is working with a rootless setup and that it doesn't mess with rootful setups. Let's add a changelog entry for this and then we can get it merged.

I think we're going to need to update the docs over [here](https://github.com/hashicorp/web-unified-docs/blob/main/content/nomad/v1.11.x/content/plugins/drivers/podman.mdx) as well. I can do that or maybe we can recruit @aimeeu if she has a moment. I'll take a look tomorrow.

aimeeu:
@tgross I can do the docs update. When will this project have its next release?

Issue #390, opened 2024-11-26
Accessing devices in rootless containers with `--group-add keep-groups`


cycneuramus:
As of Podman 3.2, there is the feature `--group-add keep-groups` which lets the container process maintain the groups of its parent process. This is useful for the case where one wants to access (and not simply mount) devices inside a rootless container, as explained in Red Hat's blog post [Using files and devices in Podman rootless containers ](https://www.redhat.com/en/blog/files-devices-podman).

I feel like I might be missing something, but it seems that accessing devices in rootless containers is currently not possible using this driver, since as far as I can tell, there is no way of passing the `keep-groups` option. So, for instance, if one would like to run something like Home Assistant (passing a Zigbee USB device) or Jellyfin (passing a video device for hardware acceleration), one would need to use rootful containers instead of the recommended Podman solution of adding the rootless user to the respective device groups and passing `--group-add keep-groups`.

PR #487, opened 2026-01-13
Add support for group_add config with keep-groups

raphlopez:
Add config and special `keep-groups` handling required to support the `--group-add=group|keep-groups` equivalent from the podman CLI. The special `keep-groups` case allows users to configure the user running within the container process to inherit the groups of the host user.

Closes https://github.com/hashicorp/nomad-driver-podman/issues/390

- [X] If a change needs to be reverted, we will roll out an update to the code within 7 days.

Job submissions can now specify arbitrary host groups that the container user should be added to, which weakens the container isolation boundary. Main use case for this is trusted workloads that need access to host resources for e.g. telemetry collection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## UNRELEASED
 
+* config: Add support for group_add config with keep-groups ([GH-487](https://github.com/hashicorp/nomad-driver-podman/pull/487))
+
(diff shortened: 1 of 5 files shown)


Harxyz:
hi, when is this expected to be released? thanks in advance.

PR #502, opened 2026-04-04
feat: Support storage_opt like docker driver

skoppe:
Taking inspiration from https://github.com/hashicorp/nomad/pull/4908 this MR adds support for specifying a storage_opt block in the Nomad job.

Available storage options can be found here
https://manpages.debian.org/experimental/containers-storage/containers-storage.conf.5.en.html

Note some options are only available on certain backing stores. For example, the `size` option is not supported on ext4 by Podman yet, even though the Linux kernel nowadays supports it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -532,6 +532,16 @@ config {
 }
 ```
 
+* **storage_opt** - (Optional) A key-value map of storage options set on container start.
+
+```hcl
+config {
+  storage_opt = {
+    size = "40G"
+  }
+}
+```
+
 * **ulimit** - (Optional) A key-value map of ulimit configurations to set to the containers to start.
 
 ```hcl
--- a/api/structs.go
+++ b/api/structs.go
@@ -222,6 +222,9 @@ type ContainerStorageConfig struct {
 	// Devices are devices that will be added to the container.
 	// Optional.
 	Devices []spec.LinuxDevice `json:"devices,omitempty"`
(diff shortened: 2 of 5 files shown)


skoppe:
Sorry to ping you @jrasell, but would you be able to assign a reviewer?

tehut (changes requested):
Overall this looks great, thanks for the submission!

I spent some time trying to validate it and I feel confident that the plumbing is right. But I ran out of time troubleshooting issues with the flag itself. 

Would you mind adding a sample jobspec and any other required config to your description?  I am pretty confident that most people who want to use this flag will be familiar with it but it's good to have documented and I'd like to get this running before I approve the PR.


I think that and a changelog entry are all we'll need to have before merging

skoppe:
Thank you for reviewing!

> Would you mind adding a sample jobspec and any other required config to your description? I am pretty confident that most people who want to use this flag will be familiar with it but it's good to have documented and I'd like to get this running before I approve the PR.
> 
> I think that and a changelog entry are all we'll need to have before merging

Will do, but not for 1 or 2 weeks due to holiday.

skoppe:
I picked this up again and during a second round of testing I noticed that podman doesn't clean up xfs project quota's. Issue is [here](https://github.com/containers/container-libs/pull/886).

Would like to get that fixed upstream before merging this unfortunately.

PR #506, opened 2026-05-13
ci: declare permissions on jira-sync and tests

arpitjain099:
Pins the default `GITHUB_TOKEN` on the two workflows that were inheriting org defaults:

- **jira-sync.yml** — `issues: read` + `contents: read`. The Jira mirror reads issue/comment metadata and writes everything to Jira via `JIRA_API_TOKEN` (gajira-login / gh-action-jira-create / gajira-transition / etc). The GitHub token doesn't need any write scopes.
- **tests.yml** — `contents: read`. Plain CI run that checks out source and runs the test suite.

YAML validated with `yaml.safe_load` on each edited file.

### DIFF
--- a/.github/workflows/jira-sync.yml
+++ b/.github/workflows/jira-sync.yml
@@ -10,6 +10,12 @@ on:
 
 name: Jira Issue Sync
 
+# Jira mirror only reads issue/comment metadata; all writes go to Jira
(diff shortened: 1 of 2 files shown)


tgross:
Closing as per https://github.com/hashicorp/nomad/pull/27959#issuecomment-4441309901

PR #507, opened 2026-05-26
fixed static IP assignment

### LINKED ISSUE #490: Static ip address not assigned
According to https://github.com/hashicorp/nomad-driver-podman/pull/445 static ip addresses shall be supported, but it seems not to be the case.

Here's my job config:

```hcl
job "nginx-podman-job" {
  datacenters = ["dc1"]
  type = "service"

  group "nginx-group" {
    count = 1

    volume "hugoconfig" {
      type      = "host"
      read_only = true
      source    = "hugoconfig"
    }

    network {
      port "http" {
        to = "8080"
      }
    }

    task "nginx-task" {
      driver = "podman"

      volume_mount {
        volume      = "hugoconfig"
        destination = "/usr/local/etc"
        read_only   = true
      }

      config {
        image = "codeberg.org/scip/freebsd/nginx:1.28.0_3_3"
        ports = ["http"]
        readonly_rootfs = false
        static_ips = ["2a01:4f8:c2c:cd3c:fefe::beef"]
      }

      resources {
        cpu = 500
        memory = 256
      }
    }
  }
}
```

When I run this with `nomad run nginx.nomad` it's being executed and running after a couple seconds without any error whatsoever.

However, when I look into the jail, the static ip address is not configured:

```console
eth0: flags=1008843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST,LOWER_UP> metric 0 mtu 1500
        options=8<VLAN_MTU>
        ether 02:44:90:c0:d9:0b
        inet 10.89.1.3 netmask 0xffffff00 broadcast 10.89.1.255
        inet6 2a01:4f8:c2c:cd3c:fefe::3 prefixlen 80
[..]
```

This is the default podman network being used:

```json
[
     {
          "name": "localv6",
          "id": "214266c302ca571f9058fcffbe3b08cc56c550806b5ba9bef01ce1da2653dae9",
          "driver": "bridge",
          "network_interface": "bridge0",
          "created": "2026-01-15T15:21:02.684574+01:00",
          "subnets": [
               {
                    "subnet": "2a01:4f8:c2c:cd3c:fefe::/80",
                    "gateway": "2a01:4f8:c2c:cd3c:fefe::1"
               },
               {
                    "subnet": "10.89.1.0/24",
                    "gateway": "10.89.1.1"
               }
          ],
          "ipv6_enabled": true,
          "internal": false,
          "dns_enabled": false,
          "ipam_options": {
               "driver": "host-local"
          },
          "containers": {
               "86d5086ed7dda5c947ea01330f8e077bc8eb2b5644323c57fd79d73976ac499c": {
                    "name": "nginx-task-4094995a-dbae-249f-251d-011a1a51a835",
                    "interfaces": {
                         "eth0": {
                              "subnets": [
                                   {
                                        "ipnet": "2a01:4f8:c2c:cd3c:fefe::3/80",
                                        "gateway": "2a01:4f8:c2c:cd3c:fefe::1"
                                   },
                                   {
                                        "ipnet": "10.89.1.3/24",
                                        "gateway": "10.89.1.1"
                                   }
                              ],
                              "mac_address": "02:44:90:c0:d9:0b"
                         }
                    }
               }
          }
     }
]
```

So, podman took just one ip address from the `/80` pool, because obviously nomad-driver-podman didn't tell it to use a static one.

My environment:

- FreeBSD 14.3-RELEASE-p7
- Nomad 1.11.1
- nomad-driver-podman 0.6.3
- Podman 5.6.1

My nomad config can be seen in #489 (first post).

**edited to add**: I also tried `args = ["--ip6=2a01:4f8:c2c:cd3c:fefe::beef"]` in the `config` block of the task. I could start the job, it runs, but the arg is ignored as well.

ritesh-harihar:
**_Summary_**

Static IP configuration `(static_ips, ipv4_address, ipv6_address, static_mac)` was silently ignored on Podman. Containers started successfully but received DHCP addresses instead of the configured static IPs.

**_Root Cause_**

The Networks field in `ContainerNetworkConfig` had an incorrect JSON tag (`"newNetworks"`) that doesn't match Podman's API. Podman expects the field to be serialized as "Networks". The mismatched key was silently discarded by Podman, so containers started without error but without static IP assignment.

Additionally, with the JSON tag fixed and `createOpts.Networks` actually reaching Podman, two latent issues in the driver became reachable: the map key was being derived from `network_mode` (which holds modes like slirp4netns or task:foo, not network names), and IP resolution for service discovery relied solely on the top-level IPAddress field, which Podman leaves empty for the per-network map case.

**_What This Fixes_**

- JSON serialization — Corrected the JSON tag so static IP config actually reaches Podman
- Network key mapping — Static IPs are mapped to the "default" network only when NSMode == api.Bridge; non-bridge modes no longer touch createOpts.Networks
- IP resolution for service discovery — driverNet.IP now falls back to the per-network map when Podman leaves the top-level IPAddress empty

**Examples**
**Before (broken):**

```
# Job config
config {
  image      = "alpine:latest"
  static_ips = ["10.88.0.100"]
}
```
```
# Result: container gets random DHCP IP
$ podman inspect <cid> --format '{{.NetworkSettings.Networks.podman.IPAddress}}'
10.88.0.5  ← wrong, should be 10.88.0.100
```

**After (fixed):**

```
$ podman inspect <cid> --format '{{.NetworkSettings.Networks.podman.IPAddress}}'
10.88.0.100  ← correct
```

Testing:

<details>

- Before fix:

`2026-05-27T17:41:26.853+0530 [INFO]  client.driver_mgr.nomad-driver-podman: Completely started container: driver=podman @module=podman container=926a40249e10c2926dd0f37dbbc58057927adfc061c979d019b51793c62e705b ip=10.88.0.3 taskID=c05a4c3c-8c44-7394-45f8-70d7057a1e43/alpine/6905e387 timestamp="2026-05-27T17:41:26.853+0530"`

- After Fix:

`2026-05-27T17:44:19.102+0530 [INFO]  client.driver_mgr.nomad-driver-podman: Completely started container: driver=podman @module=podman container=1da12a76e6a93a565f089d10943cbd87769025188daa9893911099555fb998b7 ip=10.88.0.100 taskID=fbe4886a-e638-6468-e086-b7ecbcdca30c/alpine/d7c3fe6a timestamp="2026-05-27T17:44:19.102+0530"`

  
</details>


api/structs_test.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"encoding/json"
+	"net"
+	"strings"
+	"testing"
+
+	"github.com/shoenig/test/must"
+)
+
+func TestSpecGenerator_NetworksJSONSerialization(t *testing.T) {
```
tgross: Isn't this test just exercising the stdlib's JSON serialization? Is this really a good expediture of lines-of-code and test time that's going to have to execute forever after?

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 IMPROVEMENTS:
 * config: Adds support for plugin configuration `networking.default_rootless_mode` to override the default networking mode when run rootlessly. [[GH-495](https://github.com/hashicorp/nomad-driver-podman/pull/495)]
 
+BUG FIXES:
+
```
tgross: Style nit-pick to match other changelogs:
```suggestion
```

But your changelog is also just incomplete, isn't it? Your PR description has _4_ different bugs (which again, is far too many for a single PR!) and it seems like at least a couple of them are actually different user-facing behaviors. All of those should be reflected here.

driver.go
```diff
@@ -832,7 +832,8 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 			createOpts.ContainerNetworkConfig.NetNS.NSMode = api.FromContainer
 			createOpts.ContainerNetworkConfig.NetNS.Value = BuildContainerNameForTask(otherTaskName, cfg)
 		default:
-			return nil, nil, fmt.Errorf("Unknown/Unsupported network mode: %s", podmanTaskConfig.NetworkMode)
+			// Treat any other value as a custom network name (e.g., "localv6")
```
tgross: This comment doesn't agree with the code or is at least incomplete. If the user gave us some custom network name, why would we assign "bridge" mode?

Not returning an error here also disagrees with the behavior described in the docs.

driver.go
```diff
@@ -892,7 +893,11 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 					}
 				}
 			}
-			createOpts.Networks = map[string]api.PerNetworkOptions{"default": netOpts}
+			networkName := "default"
+			if podmanTaskConfig.NetworkMode != "" && podmanTaskConfig.NetworkMode != "bridge" {
+				networkName = podmanTaskConfig.NetworkMode
+			}
+			createOpts.Networks = map[string]api.PerNetworkOptions{networkName: netOpts}
```
tgross: So the network _name_ will now be something like `slirp4netns` or "task:foo" if `task.config.network_mode` is set at all?

driver.go
```diff
@@ -1924,3 +1934,23 @@ func parseIDMapping(idmapConfig string) (*api.IDMap, error) {
...
+	if networkSettings == nil {
+		return ""
+	}
+	if networkSettings.IPAddress != "" {
+		return networkSettings.IPAddress
+	}
+	networkName := "default"
+	if networkMode != "" && networkMode != "bridge" {
+		networkName = networkMode
+	}
```
tgross: Hasn't the network name already been set in the caller earlier? (It is an awfully long function though, maybe easy to miss if there's a case where it won't have been.)

tgross (changes requested):
@ritesh-harihar in addition to my comments, can you confirm that this has been tested end-to-end with the default behavior where `group.network` takes precedence?

api/structs_test.go
ritesh-harihar: Yeah, removed it. The struct tag is exercised end-to-end by the integration path.

driver.go
ritesh-harihar: Agreed on the points mentioned. Reverted the default case to return the original error. I think Custom CNI network support  deserves its own PR with a proper docs update. I'll file a follow-up issue.

driver.go
ritesh-harihar: Understood, the map key was being derived from network_mode, which is a mode (host, slirp4netns, task:foo), not a network name.

Added a fix by gating the assignment behind `NSMode == api.Bridge` and hardcoding the key to "default". Non-bridge modes don't touch `createOpts.Networks `(Podman ignores per-network options for them). Combined with the default-case revert, only "default" can appear as the key.

driver.go
ritesh-harihar: Noted. `StartTask() `now no longer derives networkName,  it just passes the literal "default" to `resolveContainerIP`.

ritesh-harihar:
> @ritesh-harihar in addition to my comments, can you confirm that this has been tested end-to-end with the default behavior where `group.network` takes precedence?

Yes verified in a VM with Podman 5.4.2 + Nomad dev agent.

Test job: group { network { mode = "bridge" } } and a task-level static_ips = ["10.88.99.99"] (deliberately set to confirm it's ignored).

Result:

- Task event: Started — Task started by client (no errors).
- Container IP: 172.26.64.6 (CNI-assigned), not 10.88.99.99.
- No Podman error about "networks and static ip can only be used with Bridge mode networking".

The `NSMode == api.Bridge` guard skips populating `createOpts.Networks` whenever Nomad owns the network namespace(group.network, host, none, etc.), so Podman never sees the task-level static IP and the group network behavior is unchanged.

static_ips / ipv4_address / static_mac are now honored only on task-level bridge mode — which matches the pre-existing intent and Podman's own constraint.

tgross (approved):
LGTM!

PR #509, opened 2026-06-01
added support for custom Podman network names 

ritesh-harihar:
Fixes: [#1518](https://hashicorp.atlassian.net/browse/NMD-1518)

**_Summary_**

Adds support for using pre-existing Podman networks (created with podman network create) directly via network_mode. Previously, any value that wasn't a known keyword (bridge, host, none, slirp4netns, pasta, container:*, task:*, ns:*) would cause Unknown/Unsupported network mode error.

```
config {
  image        = "alpine:latest"
  network_mode = "mynet"       # ← now works
  static_ips   = ["10.89.2.50"] # ← optional
}
```

**_Root Cause_**

The default: case in the network mode switch returned an unconditional error:

```
default:
    return nil, nil, fmt.Errorf("Unknown/Unsupported network mode: %s", podmanTaskConfig.NetworkMode)
```

There was no path to attach containers to user-created networks. Additionally, even after [PR #507](https://github.com/hashicorp/nomad-driver-podman/pull/507) fixed static IPs, the Networks map was only populated inside the static IP block — meaning custom networks without static IPs silently fell back to the default bridge.

**Previous Flow vs After Flow**
Step | Before (Broken) | After (Fixed)
-- | -- | --
User sets network_mode = "mynet" | Hits default: case | Hits default: case
Invalid network name | Generic "Unknown/Unsupported" error | Clear: network "X" not found, verify with: podman network ls
NSMode assignment |  Never reached | Set to Bridge (custom networks are bridged networks)
Networks map |  Never set |  {"mynet": {}} — tells Podman which network to join
Container placement |  Error, container never starts | Container attached to correct network
Custom network + static IP | Error | Networks map overwritten with {"mynet": {static_ips: [...]}}
Custom network without static IP | Error | Networks: {"mynet": {}} — gets DHCP IP on correct network
IP resolution (driverNet.IP) | N/A | resolveContainerIP() checks Networks["mynet"].IPAddress

**_Major Changes_**
- driver.go ->  default: case validates network, sets NSMode=Bridge, populates Networks map, uses dynamic networkName for IP resolution
- network_exists.go -> New — NetworkExists() calls /v4.0.0/libpod/networks/{name}/exists

**Testing:**
<details>

1. test-custom-network.nomad
```
==> 2026-06-01T15:11:09+05:30: Monitoring deployment "6284b914"
  ✓ Deployment "6284b914" successful
    
    2026-06-01T15:11:09+05:30
    ID          = 6284b914
    Job ID      = test-custom-network
    Job Version = 0
    Status      = successful
    Description = Deployment completed successfully
    
    Deployed
    Task Group  Desired  Placed  Healthy  Unhealthy  Progress Deadline
    test        1        1       1        0
    2026-06-01T15:20:26+05:30
=== PROOF: Network name must be testnet6, NOT podman ===

Network: testnet6 | IPv4: 10.89.2.8
```

2. test-custom-network-static.nomad
```
==> 2026-06-01T15:12:00+05:30: Monitoring deployment "dfbd856d"
  ✓ Deployment "dfbd856d" successful
    
    2026-06-01T15:12:11+05:30
    ID          = dfbd856d
    Job ID      = test-custom-network-static
    Job Version = 0
    Status      = successful
    Description = Deployment completed successfully
    
    Deployed
    Task Group  Desired  Placed  Healthy  Unhealthy  Progress Deadline
    test        1        1       1        0
    2026-06-01T15:22:09+05:30
=== PROOF: Must be testnet6 with EXACT IP 10.89.2.50 ===
Network: testnet6 | IPv4: 10.89.2.50
```

3. test-custom-network-bad.nomad
```
2026-06-01T15:21:44+05:30  Not Restarting  Error was unrecoverable
2026-06-01T15:21:44+05:30  Driver Failure  rpc error: code = Unknown desc = network "does-not-exist" not found, verify with: podman network ls
```

Test Job | Expected | Result
-- | -- | --
network_mode = "testnet6" (no static IP) | Container on testnet6 | ✅ Network: testnet6 \| IPv4: 10.89.2.8
network_mode = "testnet6" + static_ips = ["10.89.2.50"] | Exact IP on testnet6 | ✅ Network: testnet6 \| IPv4: 10.89.2.50
network_mode = "does-not-exist" | Driver Failure | ✅ network "does-not-exist" not found, verify with: podman network ls

</details>

### DIFF
--- a/api/network_exists.go
+++ b/api/network_exists.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/url"
 )
 
 // NetworkExists checks if a Podman network with the given name exists.
@@ -15,7 +16,7 @@ import (
 // Unlike other endpoints that work with /v1.0.0, this one requires /v4.0.0+.
 func (c *API) NetworkExists(ctx context.Context, name string) (bool, error) {
 
-	res, err := c.Get(ctx, fmt.Sprintf("/v4.0.0/libpod/networks/%s/exists", name))
+	res, err := c.Get(ctx, fmt.Sprintf("/v4.0.0/libpod/networks/%s/exists", url.PathEscape(name)))
 	if err != nil {
 		return false, err
 	}
--- a/api/network_exists.go
+++ b/api/network_exists.go
@@ -1,0 +1,33 @@
+// Copyright IBM Corp. 2019, 2025
+// SPDX-License-Identifier: MPL-2.0
+
+package api
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+// NetworkExists checks if a Podman network with the given name exists.
+// Returns true if the network exists (HTTP 204), false if not found (HTTP 404).
+// Note: The /networks/{name}/exists endpoint was introduced in Podman API v4.
+// Unlike other endpoints that work with /v1.0.0, this one requires /v4.0.0+.
+func (c *API) NetworkExists(ctx context.Context, name string) (bool, error) {
+
+	res, err := c.Get(ctx, fmt.Sprintf("/v4.0.0/libpod/networks/%s/exists", name))
+	if err != nil {
+		return false, err
+	}
+
+	defer ignoreClose(res.Body)
+
+	switch res.StatusCode {
+	case http.StatusNoContent:
+		return true, nil
+	case http.StatusNotFound:
+		return false, nil
+	default:
+		return false, fmt.Errorf("unexpected status code checking network %q: %d", name, res.StatusCode)
+	}
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 IMPROVEMENTS:
 * config: Adds support for plugin configuration `networking.default_rootless_mode` to override the default networking mode when run rootlessly. [[GH-495](https://github.com/hashicorp/nomad-driver-podman/pull/495)]
-* config: Support custom Podman network names for static IP/MAC options via `network_mode`. [[GH-TBD](https://github.com/hashicorp/nomad-driver-podman/pull/TBD)]
+* config: Support custom Podman network names for static IP/MAC options via `network_mode`. [[GH-509](https://github.com/hashicorp/nomad-driver-podman/pull/509)]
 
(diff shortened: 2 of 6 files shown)


ritesh-harihar:
Added a `NetworkExists()` API call before container creation.

- Without pre-flight: Podman returns a generic error at ContainerCreate time — opaque to the user and hard to distinguish from other creation failures.
- With pre-flight: 
1) Catches typos early — A typo in network_mode (e.g., "testnt6" instead of "testnet6") would otherwise pass through the driver silently and fail deep inside Podman's container create with an opaque error. The pre-flight check surfaces it as a clear "not found" at the driver level.
2) Fail fast — We catch the problem before building the full container spec, downloading images, or setting up the task directory. Cheaper failure.

If you think the UX improvement doesn't justify the extra call, I can remove NetworkExists() entirely and let Podman's ContainerCreate handle the error.

api/network_exists.go
```diff
@@ -0,0 +1,33 @@
...
+	"net/http"
+)
+
+// NetworkExists checks if a Podman network with the given name exists.
+// Returns true if the network exists (HTTP 204), false if not found (HTTP 404).
+// Note: The /networks/{name}/exists endpoint was introduced in Podman API v4.
+// Unlike other endpoints that work with /v1.0.0, this one requires /v4.0.0+.
+func (c *API) NetworkExists(ctx context.Context, name string) (bool, error) {
+
+	res, err := c.Get(ctx, fmt.Sprintf("/v4.0.0/libpod/networks/%s/exists", name))
```
schmichael: We should probably have called `url.PathEscape(name)` every single time we interpolate user input (a `network_mode` parameter). It seems unlikely this could be used as a vulnerability, but escaping user input is a good practice. (And just because *I* can't think of an attack doesn't mean someone more clever can't!)

```suggestion
	res, err := c.Get(ctx, fmt.Sprintf("/v4.0.0/libpod/networks/%s/exists", url.PathEscape(name)))
```

We shouldn't fix unrelated instances of input path escaping in this PR, but it's probably worth doing once to establish the best practice before someone finds an issue.

schmichael (dismissed):

schmichael (approved):

PR #511, opened 2026-06-05
Fix: userns=auto resulting in no UID/GID mapping

### LINKED ISSUE #503: userns=auto results in no mapping
I'm attempting to set `userns=auto` in a rootful configuration. If I just set that, it fails silently, but if I do `userns = "auto:size=65536"`, then it works.

I got the idea to run some kind of configuration from https://github.com/hashicorp/nomad-driver-podman/issues/470.

ritesh-harihar:
Fixes: https://github.com/hashicorp/nomad-driver-podman/issues/503
[JIRA Ticket](https://hashicorp.atlassian.net/browse/NMD-1415)

**_Summary_**

Setting `userns = "auto"` in a rootful Podman task configuration silently fails to apply user namespace isolation. The container runs as actual root on the host `(uid_map: 0 0 4294967295)`, defeating the purpose of auto user namespace isolation.

However, `userns = "auto:size=65536"` works correctly — adding any option after the colon triggers proper behavior. It took a different code path.

**_Root Cause_**

**1) Early return skips mapping initialization for plain `auto`**

```
if len(modeWithConfig) == 1 {
    return api.Namespace{NSMode: mode}, nil, nil  // ← nil mappings!
}
```
When `userns="auto"` has no colon, the parser returns early with nil IDMappings. The case "auto" branch that sets AutoUserNs: true is never reached. Podman receives the auto mode flag but no mapping instructions, so it does nothing.

**2) Error variable mismatch swallows parse failures**
```
userns, mappings, userNSerr := parseUserNSConfig(...)
if err != nil {  // ← wrong variable! should be userNSerr
```

**_Fix_**

Ensure that even bare `userns = "auto"` (without sub-options) returns a properly initialized `&IDMappingOptions{AutoUserNs: true, UIDMap: []IDMap{}, GIDMap: []IDMap{}}`. The mappings struct is now constructed before the sub-options check, so both `"auto"` and `"auto:size=N,..."` share the same initialization path.

**_Before/After Flow_**

Scenario |   | IDMappings sent to Podman | Container uid_map
-- | -- | -- | --
userns = "auto" (rootful) | Before | nil (omitted from JSON) | 0 0 4294967295 (no isolation)
  | After | &{AutoUserNs: true, UIDMap: [], GIDMap: []} | 0 300000 1024 (isolated)
userns = "auto:size=65536" | Before | &{AutoUserNs: true, Size: 65536} | 0 300000 65536 ✓
  | After | &{AutoUserNs: true, Size: 65536} | 0 300000 65536 ✓
userns = "auto:invalid=x" (error handling) | Before | (error swallowed silently) | Container starts without isolation
  | After | (error returned to caller) | Task fails with "failed to parse userns configuration" ✓

**_Testing:_**
<details>

**1) Previous Flow:**  
```
2026-06-05T12:54:21.609665941+05:30 stdout F          0          0 4294967295
```
UID 0 in the container maps to UID 0 on the host, no user namespace isolation at all. The container is running as actual root. 

**2) Current Flow**
```
2026-06-05T12:59:44.920613444+05:30 stdout F          0     300000       1024
```

UID 0 in the container is actually UID 300000 on the host. The container only has access to 1024 UIDs (300000–301023). Proper isolation. Even if a process escapes the container, it's just an unprivileged user (300000) on the host — not root.

`userns = "auto"` now properly tells Podman to allocate automatic UID/GID mappings from the configured subuid/subgid ranges, providing user namespace isolation.

</details>

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -1877,7 +1877,7 @@ func parseUserNSConfig(userNSConfig string) (api.Namespace, *api.IDMappingOption
 	modeWithConfig := strings.SplitN(userNSConfig, ":", 2)
 	mode := api.NamespaceMode(modeWithConfig[0])
 
-	config := ""
+	var config string
 	if len(modeWithConfig) == 2 {
 		config = modeWithConfig[1]
 	}
--- a/driver.go
+++ b/driver.go
@@ -739,7 +739,7 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 	// Populate --userns mode only if configured
 	if podmanTaskConfig.UserNS != "" {
 		userns, mappings, userNSerr := parseUserNSConfig(podmanTaskConfig.UserNS)
-		if err != nil {
+		if userNSerr != nil {
 			return nil, nil, fmt.Errorf("failed to parse userns configuration: %w", userNSerr)
 		}
 		createOpts.ContainerSecurityConfig.UserNS = userns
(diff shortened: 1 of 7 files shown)


driver.go
```diff
@@ -1877,29 +1877,32 @@ func parseUserNSConfig(userNSConfig string) (api.Namespace, *api.IDMappingOption
 	modeWithConfig := strings.SplitN(userNSConfig, ":", 2)
 	mode := api.NamespaceMode(modeWithConfig[0])
 
-	if len(modeWithConfig) == 1 {
-		// if there's no additional configuration, we can bail out early
-		return api.Namespace{NSMode: mode}, nil, nil
+	config := ""
```
tehut: nit: this is just stylistic but I find it clearer to use `var config string` if I'm intentionally setting a var to the zero value.

driver_test.go
```diff
@@ -2890,6 +2904,28 @@ func TestUserNSConfigParsing(t *testing.T) {
 
 }
 
+func TestPodmanDriver_StartTask_InvalidUserNS_ReturnsParseError(t *testing.T) {
```
tehut: I don't think we really need this test case since it's really just testing the wrapper around parseUserNSConfig and you've added the case the catch this bug to that test

tehut (dismissed):
I left two small suggestions but they're not really blocking.  I would like to see that unnecessary test removed especially since it's pretty heavy but if you wanted to do that in a separate PR, I'm fine with it.

tehut (approved):
🚢 🚢 🚢

PR #527, opened 2026-07-29
security: prevent symlink escape in volumes block

gulducat:
Guards against a similar issue in the docker driver: https://github.com/hashicorp/nomad/commit/db221a2e1f4142403a41a414705a5b52789d79a1

Fortunately, this is more defense-in-depth compared to the docker case (it was vulnerable by default):

- podman driver plugin config must have `volumes.enabled = false` to explicitly disable the `volumes` driver option
- this change now fails the task before ever mounting the volume
- without it, the volume would be mounted, but selinux would prevent accessing the file (unless selinux is disabled/bypassed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## UNRELEASED
 
+SECURITY:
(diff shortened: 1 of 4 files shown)


chrisroberts (dismissed):
Including a test to exercise the check would be nice to include, but this LGTM :+1:

Issue #182, opened 2022-07-21
Nomad does not see stopping jobs with podman 4.1.1 (rest-api response change)


DemonicTutor:
nomad 1.3.2
nomad-podman-driver 0.4.0
podman 4.1.1 (upgraded from 4.1.0)
CoreOS 36.20220703.3.1 (https://getfedora.org/en/coreos?stream=stable#) Release Date: Jul 18, 2022
aarch64

When stopping a job via Nomad the containers do get stopped but not removed (gc=true is configured) and the job status in nomad never changes to stopped.
When restarting the new allocation stays in *starting* and previous in *running* (but the first container gets stopped which matches our configs)

I think this  [commit](https://github.com/containers/podman/pull/14580/files) in podman changed the rest-api stats response from HTTP 404 NOT FOUND to 200 OK if a container has been stopped and this seems to break the driver api 
[nomad-driver-podman/container_stats.go at 50007e0c40702d24c9710a8030a72a875826e743 · hashicorp/nomad-driver-podman](https://github.com/hashicorp/nomad-driver-podman/blob/50007e0c40702d24c9710a8030a72a875826e743/api/container_stats.go#L26-L28) which expects 404 to handle it as *gone*

podman via console:
```
[markus@mephistopheles ~]$ ssh 0.158.3.76
Fedora CoreOS 36.20220703.3.1
Tracker: https://github.com/coreos/fedora-coreos-tracker
Discuss: https://discussion.fedoraproject.org/tag/coreos
[xxx@ip-10-158-3-76 ~]$ sudo podman ps -a
CONTAINER ID  IMAGE                                 COMMAND               CREATED      STATUS                       PORTS       NAMES
94de9d4d1bc7  [xxxx)       -config.file /dat...  3 hours ago  Up 3 hours ago                           promtail
97b88327af7d  [xxxx)  --path.procfs /ho...  3 hours ago  Up 3 hours ago                           prom-ne
7438328ead17  [xxxx)         agent -retry-join...  3 hours ago  Up 3 hours ago                           consul
97992073a31a  [xxxx)          agent -config=/no...  3 hours ago  Up 3 hours ago                           nomad
96343dff92bf  [xxxx)        traefik               3 hours ago  Up 3 hours ago                           traefik
2104ed346090  [xxxx)                             3 hours ago  Exited (143) 57 minutes ago              rng-9d62b66a-8b38-b1d9-5181-9fc4db8b5a27
[...]

PR #530, opened 2026-08-26
driver: separate lifecycle monitoring from stats

hkwi:
## Summary

- use Podman's container wait endpoint as the source of truth for task exit
- keep container stats polling responsible only for resource metrics
- inspect the container whenever wait returns, including after a wait error
- reconnect the wait request when inspect still reports a non-terminal state
- use the no-timeout streaming HTTP client for blocking wait requests
- stop stats emission when the task exits

## Motivation

The current container monitor infers task exit from the per-container stats response. This couples lifecycle detection to a metrics endpoint whose stopped-container behavior has changed across Podman releases (#182, #277).

This change follows the lifecycle/metrics separation proposed by #80, but uses a per-container `wait?condition=exited` request instead of a global event stream. A wait request can be re-established after a transport failure, and inspect provides reconciliation if the request races with a short-lived container exit.

Stats errors no longer change task lifecycle state. Stats polling also waits for the first collection interval, giving the lifecycle monitor a chance to finalize containers that exit before the first metrics sample.

```mermaid
sequenceDiagram
    participant D as Nomad Podman driver
    participant W as Podman wait API
    participant S as Podman stats API
    participant I as Podman inspect API

    par Lifecycle
        D->>W: wait(condition=exited)
        W-->>D: exited or transport error
        D->>I: inspect container
        I-->>D: terminal state and exit code
        D->>D: finalize task
    and Metrics
        D->>S: stats(stream=false)
        S-->>D: metrics or error
        Note over D,S: Never changes task lifecycle state
    end
```

## Failure handling

- If wait returns and inspect reports `exited`, `stopped`, or `dead`, the task is finalized from inspect data.
- If inspect reports that the container is still active, wait is retried after a short delay.
- If wait fails after the container has already exited, inspect still finalizes the task.
- If the container no longer exists, the existing unknown-exit-code behavior is preserved.
- Driver shutdown cancels the blocking wait request through its context.

## Tests

- `go test ./... -run '^(TestContainerWait|TestContainerMonitor|TestContainerExited|TestStatsMonitor)' -count=1`
- `go test -race . -run '^(TestContainerMonitor|TestContainerExited|TestStatsMonitor)' -count=1`
- `go vet ./...`

Full Podman integration tests were not run locally because this environment does not have access to a Podman service socket. The existing GitHub Actions job provides the Podman-backed test environment.

Related: #80, #277, #278

### DIFF
--- a/api/container_wait.go
+++ b/api/container_wait.go
@@ -10,10 +10,16 @@ import (
 	"strings"
 )
 
-// ContainerWait waits on a container to met a given condition
+// ContainerWait waits on a container to meet one of the given conditions.
+// It uses the streaming client because a container may run longer than the
+// regular API client timeout.
 func (c *API) ContainerWait(ctx context.Context, name string, conditions []string) error {
-
-	res, err := c.Post(ctx, fmt.Sprintf("/v1.0.0/libpod/containers/%s/wait?condition=%s", name, strings.Join(conditions, "&condition=")), nil)
+	path := fmt.Sprintf("/v1.0.0/libpod/containers/%s/wait?condition=%s", name, strings.Join(conditions, "&condition="))
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.baseUrl+path, nil)
+	if err != nil {
+		return err
+	}
+	res, err := c.Do(req, true)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


hkwi:
podman-container-tools/podman#29656 would fix the root cause, that returned http status code 200 with error json.

